PR #44, opened 2015-04-23
Add iOS target

### LINKED ISSUE #16: Add an iOS framework target

phatblat:
Set up the iOS target build pretty much the same as ObjectiveGit.

iOS tests are not included. The `SwiftGit2-iOSTests` target is just a stub for now. It needs a library to unzip the test repos since `NSTask` is not available on iOS. This will be handled in another PR.

Resolves #16


mdiep:
Thanks for working on this! I'll try to take a look later today, but may not get to it until the weekend.

Cartfile.resolved
```diff
@@ -1,5 +1,5 @@
 github "modocache/Guanaco" "26b6ceff4b8912abd0dfc3a9eed0a46527e040e3"
 github "Quick/Nimble" "v0.4.2"
 github "Quick/Quick" "155f858bce082373fc60d0c0a63415bf336c676b"
-github "antitypical/Result" "0.3"
+github "antitypical/Result" "0.3.2"
```
mdiep: Can you also update the `Cartfile` to specify that we need a release compatible with 0.3.2?

.gitmodules
```diff
@@ -16,3 +16,9 @@
 [submodule "Carthage/Checkouts/Result"]
 	path = Carthage/Checkouts/Result
 	url = https://github.com/antitypical/Result.git
+[submodule "External/ios-openssl"]
```
mdiep: Can you change this to be `External/openssl`?

SwiftGit2.xcodeproj/project.pbxproj
```diff
@@ -295,6 +509,15 @@
...
+		F807D4A41ADCA66800A2DC7A /* SwiftGit2-iOSTests.xctest */ = {isa = PBXFileReference; explicitFileType = wrapper.cfbundle; includeInIndex = 0; path = "SwiftGit2-iOSTests.xctest"; sourceTree = BUILT_PRODUCTS_DIR; };
+		F807D5041ADCAC7C00A2DC7A /* libz.dylib */ = {isa = PBXFileReference; lastKnownFileType = "compiled.mach-o.dylib"; name = libz.dylib; path = Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS8.3.sdk/usr/lib/libz.dylib; sourceTree = DEVELOPER_DIR; };
+		F807D5061ADCAC8300A2DC7A /* libiconv.dylib */ = {isa = PBXFileReference; lastKnownFileType = "compiled.mach-o.dylib"; name = libiconv.dylib; path = Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS8.3.sdk/usr/lib/libiconv.dylib; sourceTree = DEVELOPER_DIR; };
+		F807D52F1ADCDC3B00A2DC7A /* liblibgit2-iOS.a */ = {isa = PBXFileReference; explicitFileType = archive.ar; includeInIndex = 0; path = "liblibgit2-iOS.a"; sourceTree = BUILT_PRODUCTS_DIR; };
+		F807D5471ADCDCF300A2DC7A /* script */ = {isa = PBXFileReference; lastKnownFileType = folder; path = script; sourceTree = "<group>"; };
```
mdiep: You added the `script` folder as a reference in the Xcode project. I don't believe that's needed. Can you remove it?

SwiftGit2Tests/Fixtures/Fixtures.swift
```diff
@@ -33,8 +33,14 @@ final class Fixtures {
...
 	func setUp() {
 		NSFileManager.defaultManager().createDirectoryAtURL(directoryURL, withIntermediateDirectories: true, attributes: nil, error: nil)
-		
-		let bundle = NSBundle(identifier: "org.libgit2.SwiftGit2-OSXTests")!
+
+		#if os(OSX)
+			let platform = "OSX"
+		#else
+			let platform = "iOS"
+		#endif
```
mdiep: Can you remove the indentation from these preprocessor bits? And use normal indentation for the variables?

SwiftGit2Tests/Fixtures/Fixtures.swift
```diff
@@ -47,12 +53,20 @@ final class Fixtures {
...
+            task.arguments = [ "-qq", "-d", directoryURL.path!, fileURL.path! ]
+            
+            task.launch()
+            task.waitUntilExit()
+
+        #else
+
+            assertionFailure("Tests not supported on iOS yet")
+
+        #endif
```
mdiep: Same thing here

SwiftGit2.xcodeproj/xcshareddata/xcschemes/OpenSSL-iOS.xcscheme
```diff
@@ -0,0 +1,110 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mdiep: I don't think we need all these extra schemes. Can you remove them?

phatblat:
@mdiep Thanks for the feedback! I'll clean these up this weekend.

mdiep:
Great! Thanks again for tackling this. :sparkles:

Leave a comment when you're done and I'll take another look.

phatblat:
Any thoughts on 9c7c53a97? I've limited that to `TARGET_OS_IPHONE`, but it feels like a hack which might be resolved with the correct header and/or linker paths.

mdiep:
> Any thoughts on 9c7c53a? I've limited that to TARGET_OS_IPHONE, but it feels like a hack which might be resolved with the correct header and/or linker paths.

Yeah, I think you should be able to get around that by adjusting the _Other Linker Flags_ for the iOS target. It should be some combination of removing unneeded items from that list or adding flags to point to the libraries in `External/`.

Do you get those same errors when you build the OS X target?

phatblat:
No, I only get those errors when building the iOS target. The temp fix I put in actually causes errors in the OS X target, which is why I wrapped them with `TARGET_OS_IPHONE`. I'll dig into the linker flags and remove those declarations when I find the culprit.

mdiep:
You should also verify that libssh2 is being built for x86_64. You can check which architectures it contains with `lipo`.

phatblat:
OK, I've fixed up all of the outstanding items but still need to review the architectures in the libssh2 library.

In this last round of build script revisions, I found that there are a ton of warnings like the following being emitted by the `update_libgit2_ios` script.

````
Incompatible pointer types passing 'size_t *' (aka 'unsigned long *') to parameter of type 'unsigned int *
```

The warnings seem are being generated when building for 32-bit architectures (i386 or armv7). Turns out ObjectiveGit is also generating these warnings, but they are piped to a separate log file and thus not easily discoverable. I'll dig into this and hopefully fix it for both projects.

phatblat:
```
lipo -info openssl/lib/libssl.a openssl/lib/libcrypto.a libssh2-ios/lib/libssh2-ios.a libgit2-ios/libgit2-ios.a    
Architectures in the fat file: openssl/lib/libssl.a are: i386 armv7 armv7s x86_64 arm64
Architectures in the fat file: openssl/lib/libcrypto.a are: i386 armv7 armv7s x86_64 arm64
Architectures in the fat file: libssh2-ios/lib/libssh2-ios.a are: i386 armv7 armv7s x86_64 arm64
Architectures in the fat file: libgit2-ios/libgit2-ios.a are: i386 armv7 armv7s x86_64 arm64
```

script/update_libssh2_ios
```diff
@@ -28,14 +28,14 @@ function build_ssh2 ()
...
     mkdir -p "${ROOT_PATH}/External/libssh2-ios/bin/${SDKNAME}-${ARCH}.sdk"
-    LOG="${ROOT_PATH}/External/libssh2-ios/bin/${SDKNAME}-${ARCH}.sdk/build-libssh2.log"
 
-    echo "${LOG}"
-
-    ./buildconf >> "${LOG}" 2>&1
-    ./configure --host=${HOST} --prefix="${ROOT_PATH}/External/libssh2-ios/bin/${SDKNAME}-${ARCH}.sdk" --with-openssl --with-libssl-prefix="${ROOT_PATH}/External/ios-openssl" --disable-shared --enable-static >> "${LOG}" 2>&1
-    make >> "${LOG}" 2>&1
-    make install >> "${LOG}" 2>&1
+    ./buildconf > /dev/null 2>&1
```
mdiep: Can you remove the `> /dev/null` here? I was missing some build tools, so I couldn't build. But it took some digging to figure out what was happening.

Cartfile
```diff
@@ -1 +1 @@
-github "antitypical/Result"
+github "antitypical/Result" >= 0.3.2
```
mdiep: This should really be `~> 0.3.2`, since we won't be compatible with 0.4 and later. Sorry I wasn't clearer before.

script/update_libssh2_ios
mdiep: Oh, I see you disabled this because the config info was showing up as errors. It sounds like we should (1) output this to a log and (2) print the location of that log as part of the build script.

mdiep:
I'm still getting this:

```
Undefined symbols for architecture x86_64:
  "_opendir$INODE64", referenced from:
      _git_path_direach in libgit2-ios.a(path.c.o)
      _git_path_dirload in libgit2-ios.a(path.c.o)
  "_readdir_r$INODE64", referenced from:
      _git_path_direach in libgit2-ios.a(path.c.o)
      _git_path_dirload in libgit2-ios.a(path.c.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

Any idea why that would be the case?

It may just be the configuration on my machine—not anything wrong with this PR.

SwiftGit2Tests/Fixtures/Fixtures.swift
```diff
@@ -47,12 +53,20 @@ final class Fixtures {
...
 		task.launchPath = "/usr/bin/unzip"
 		task.arguments = [ "-qq", "-d", directoryURL.path!, fileURL.path! ]
-		
+            
 		task.launch()
 		task.waitUntilExit()
+
+#else
+
+		assertionFailure("Tests not supported on iOS yet")
```
mdiep: If we're not able to do this as part of this PR, can you add a comment explaining why this is the case?

phatblat:
32573cf4 fixed those "Undefined symbols" errors about `_opendir$INODE64` for me. Not sure why besides the fact that the path removed was where the OS X built products end up. The `"$(BUILD_DIR)/$(CONFIGURATION)$(EFFECTIVE_PLATFORM_NAME)"` line should resolve to the same when building for OS X, but also include the `-iphoneos` or `-iphonesimulator` when building for iOS.

SwiftGit2Tests/Fixtures/Fixtures.swift
phatblat: I haven't had much time to research libraries for unzipping on iOS. I think it would be best to handle that in another PR. I'll revise the description to make it clear that tests aren't included.

mdiep: I'm :+1: to adding that in a separate PR. Let's just add a comment here then explaining that the fixtures can't be unzipped.

modocache: I ran into this issue with Gift as well: https://github.com/modocache/Gift/commit/f1c73dfdc3c042d2e6027bb0be321787db792511

Since `system()` and `NSTask` aren't available, I used `zlib`--or, more specifically, I used @soffes's [SSZipArchive], which wraps `zlib` with a friendlier API.

phatblat:
Superseded by #52

PR #47, opened 2016-01-16
SwiftGit2 won't build with Xcode 7.2 that uses Swift 2.0

qnoid:
Ohai there.

Before merging, two things need reviewing.

* Under [Pointers#46](https://github.com/SwiftGit2/SwiftGit2/compare/master...windmill-io:swift-2.0?expand=1#diff-bfa5542ea1df458efa0c5a47ffb7eb02R46), made an educated guess and set a default case of *GIT_OBJ_ANY* since the switch needs to be exhaustive.

* Under [Pointers#60](https://github.com/SwiftGit2/SwiftGit2/compare/master...windmill-io:swift-2.0?expand=1#diff-bfa5542ea1df458efa0c5a47ffb7eb02L60) now switching on the *git_otype* instance as opposed to its value.


mdiep:
Wow, thanks! :tada: This has been on my list of things to do, but I haven't gotten around to it. I'll take some time to review in the next few days.

qnoid:
You are welcome @mdiep :)

Can you please provide some context on the project? 
Have no prior knowledge of libgit2 and objective-c bindings.

I can see that libgit2 provides Objective-C bindings as part of its release, is it due to lack of resources, priorities that this project lies outside the main repo?

Development of Swift bindings seems stagnant. Is it due to lack of time or general interest? What motivated you to start this?

mdiep:
> I can see that libgit2 provides Objective-C bindings as part of its release, is it due to lack of resources, priorities that this project lies outside the main repo?

ObjectiveGit is only under the libgit2 organization for historical reasons. Libgit2 now prefers for bindings to live on their own—but existing bindings won't be moved. Among other reasons, this is because the development of these projects is completely separate.

> Development of Swift bindings seems stagnant. Is it due to lack of time or general interest? What motivated you to start this?

Lack of time, mostly. I have some other projects that I'm trying to further (among them, https://github.com/Carthage/Carthage).

I started this because (a) the design of ObjectiveGit is lacking and (b) as long as I was starting over, it made sense to do so in Swift.

SwiftGit2.xcworkspace/xcshareddata/SwiftGit2.xcscmblueprint
```diff
@@ -0,0 +1,107 @@
+{
```
mdiep: This file really shouldn't be checked in. Can you remove it and then add `*.xcscmblueprint` to the `.gitignore`.

SwiftGit2/OID.swift
```diff
@@ -68,7 +68,7 @@ extension OID: Hashable {
 			self.oid.id.6,
 			self.oid.id.7
 		]
-		return reduce(bytes, 0, { (hash, byte) in
+		return bytes.reduce(0, combine:{ (hash, byte) in
```
mdiep: Style-wise, this would be better with trailing closure syntax:

```swift
return bytes.reduce(0) { (hash, byte) in
	return Int(hash << 8) | Int(byte)
}
```

Would you mind adjusting it?

SwiftGit2/Repository.swift
```diff
@@ -244,15 +244,15 @@ final public class Repository {
 			}
 			.map {
 				self.referenceWithName($0)
-			}
+		}
```
mdiep: Can you undo  this indentation change?

SwiftGit2/Repository.swift
```diff
@@ -278,7 +278,7 @@ final public class Repository {
 		return referencesWithPrefix("refs/heads/")
 			.map { (refs: [ReferenceType]) in
 				return refs.map { $0 as! Branch }
-			}
+		}
```
mdiep: And this one?

SwiftGit2/Repository.swift
```diff
@@ -286,7 +286,7 @@ final public class Repository {
 		return referencesWithPrefix("refs/remotes/")
 			.map { (refs: [ReferenceType]) in
 				return refs.map { $0 as! Branch }
-			}
+		}
```
mdiep: And this one?

SwiftGit2/Repository.swift
```diff
@@ -304,7 +304,7 @@ final public class Repository {
 		return referencesWithPrefix("refs/tags/")
 			.map { (refs: [ReferenceType]) in
 				return refs.map { $0 as! TagReference }
-			}
+		}
```
mdiep: And this one?

SwiftGit2/Pointers.swift
```diff
@@ -45,27 +45,29 @@ public enum Pointer: PointerType {
...
+		case Tree(oid):
 			return GIT_OBJ_TREE
-		case let Blob(oid):
+		case Blob(oid):
 			return GIT_OBJ_BLOB
-		case let Tag(oid):
+		case Tag(oid):
 			return GIT_OBJ_TAG
+		default:
+			return GIT_OBJ_ANY
```
mdiep: This default case isn't needed. I suspect Swift is getting confused about what `Commit`, etc. mean here. Since the `let` was removed, the variables should also be removed:

```swift
	public var type: git_otype {
		switch self {
		case Commit:
			return GIT_OBJ_COMMIT
		case Tree:
			return GIT_OBJ_TREE
		case Blob:
			return GIT_OBJ_BLOB
		case Tag:
			return GIT_OBJ_TAG
		}
	}
```

mdiep:
This is great! I left some notes of things to fix up, but they're all pretty minor.

qnoid:
Thank you for the answers. Paint the picture I was looking for :)
IMHO, it is worth having these answers in the README.

Tho I'm not sure how much involvement I'm going to have, it's useful to know what you think ObjectiveGit's design is lacking. I can then use it as guidance. Are these the points you've already made under the **Design** section in the README?

Thanks for bringing SwiftGit2 to the world!

mdiep:
> Are these the points you've already made under the Design section in the README?

Yes. I wrote up some more about them in https://github.com/libgit2/objective-git/issues/407.

mdiep:
Fantastic! Thank you!

PR #50, opened 2016-02-10
Remove Box from .gitmodules

tombooth:
During the upgrade to Swift 2.0 in `5c8a600` the dependency on the
library Box was removed, but the submodule wasn't fully.

This caused projects depending on SwiftGit2 to fail when Carthage was
recursively retrieving dependencies as it would find reference to Box
but no directory for it. The error I encountered was as below:

```
Parse error: expected submodule commit SHA in output of task (ls-tree -z ... Carthage/Checkouts/Box) but encountered:
```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -16,6 +16,3 @@
 [submodule "Carthage/Checkouts/Result"]
 	path = Carthage/Checkouts/Result
 	url = https://github.com/antitypical/Result.git
-[submodule "Carthage/Checkouts/Box"]
-	path = Carthage/Checkouts/Box
-	url = https://github.com/robrix/Box.git


tombooth:
FYI this would have been caught had the CI build been using Carthage rather than just initialising the submodules. Though I am sure there is a reason for that, I've only skimmed this very quickly.

mdiep:
Thanks!

> FYI this would have been caught had the CI build been using Carthage rather than just initialising the submodules. Though I am sure there is a reason for that, I've only skimmed this very quickly.

There's no real reason for that. It's just how it was done. :relaxed:

I'd very happily accept a PR that changes the CI build to use Carthage if you're interested!

PR #53, opened 2016-02-23
Correct the dependency between openssl and libssh2

tombooth:
Previously, libgit2 was depending on both openssl and libssh2. This was
incorrect as libssh2 requires openssl for its build phase. I have moved
the openssl dependency back to libssh2.

I was testing using the scripts/cibuild executable, which masks any
dependency issues with openssl as it prebuilds it. This prebuilding was
done to get around TravisCI timeouts and so can't be removed to stop
this being missed again.


mdiep:
:+1:

PR #57, opened 2016-03-03
Fix EXC_BAD_ACCESS when progress updates

tombooth:
Previously when a block was passed into checkout() it would be copied
but then the memory would be freed as nothing had retained a reference
to the copy. This meant that when block() was called in
SG2CheckoutProgressCallback() it would crash due to an attempt to
execute a random chunk of memory that previously contained the block.

This code now retains the payload after copying it.


mdiep:
Good catch! Thanks.

PR #58, opened 2016-03-03
Clone

### LINKED ISSUE #24: API to clone a repository

tombooth:
This implements a clone function and credential management so a user of this library can clone public and private repositories.

This function leans on https://github.com/libgit2/objective-git/blob/9b092aceb2bdebec5a1ad14d6166900b783da521/ObjectiveGit/GTRepository.m#L214 for some of its implementation detail.

I still expect the tests to fail as I am having some issues with what seems to be DNS resolution in the iOS scheme running on TravisCI.

Closes #24

### DIFF
--- a/SwiftGit2/Credentials.swift
+++ b/SwiftGit2/Credentials.swift
@@ -1,0 +1,50 @@
+//
+//  Credentials.swift
+//  SwiftGit2
+//
+//  Created by Tom Booth on 29/02/2016.
+//  Copyright © 2016 GitHub, Inc. All rights reserved.
+//
+
+import Foundation
+import Result
+
+private class Wrapper<T> {
+	let value: T
+
+	init(_ value: T) {
+		self.value = value
+	}
+}
+
+public enum Credentials {
+	case Default()
+	case Plaintext(username: String, password: String)
+	case SSHMemory(username: String, publicKey: String, privateKey: String, passphrase: String)
+
+	internal static func fromPointer(pointer: UnsafeMutablePointer<()>) -> Credentials {
+		return Unmanaged<Wrapper<Credentials>>.fromOpaque(COpaquePointer(pointer)).takeRetainedValue().value
+	}
+
+	internal func toPointer() -> UnsafeMutablePointer<()> {
+		return UnsafeMutablePointer(Unmanaged.passRetained(Wrapper(self)).toOpaque())
+	}
+}
+
+/// Handle the request of credentials, passing through to a wrapped block after converting the arguments.
+/// Converts the result to the correct error code required by libgit2 (0 = success, 1 = rejected setting creds, -1 error)
+internal func credentialsCallback(cred: UnsafeMutablePointer<UnsafeMutablePointer<git_cred>>, _: UnsafePointer<Int8>, _: UnsafePointer<Int8>, _: UInt32,
+	payload: UnsafeMutablePointer<()>) -> Int32 {
+	let result: Int32
+
+	switch Credentials.fromPointer(payload) {
+	case .Default():
+		result = git_cred_default_new(cred)
+	case .Plaintext(let username, let password):
+		result = git_cred_userpass_plaintext_new(cred, username, password)
+	case .SSHMemory(let username, let publicKey, let privateKey, let passphrase):
+		result = git_cred_ssh_key_memory_new(cred, username, publicKey, privateKey, passphrase)
+	}
+
+	return (result != GIT_OK.rawValue) ? -1 : 0
+}
--- a/SwiftGit2/Repository.swift
+++ b/SwiftGit2/Repository.swift
@@ -27,16 +27,19 @@ private func checkoutProgressCallback(path: UnsafePointer<Int8>, completed_steps
 	}
 }
 
-/// Helper function for initializing libgit2 got_checkout_options.
+/// Helper function for initializing libgit2 git_checkout_options.
 ///
+/// :param: strategy The strategy to be used when checking out the repo, see CheckoutStrategy
 /// :param: progress A block that's called with the progress of the checkout.
 /// :returns: Returns a git_checkout_options struct with the progress members set.
-private func checkoutOptions(progress: CheckoutProgressBlock? = nil) -> git_checkout_options {
+private func checkoutOptions(strategy: CheckoutStrategy, progress: CheckoutProgressBlock? = nil) -> git_checkout_options {
 	// Do this because GIT_CHECKOUT_OPTIONS_INIT is unavailable in swift
 	let pointer = UnsafeMutablePointer<git_checkout_options>.alloc(1)
 	git_checkout_init_options(pointer, UInt32(GIT_CHECKOUT_OPTIONS_VERSION))
 	var options = pointer.move()
 	pointer.dealloc(1)
+
+	options.checkout_strategy = strategy.git_checkout_strategy.rawValue
 
 	if progress != nil {
 		options.progress_cb = checkoutProgressCallback
@@ -48,9 +51,50 @@ private func checkoutOptions(progress: CheckoutProgressBlock? = nil) -> git_chec
 	return options
 }
 
+private func fetchOptions(credentials: Credentials) -> git_fetch_options {
+	let pointer = UnsafeMutablePointer<git_fetch_options>.alloc(1)
+	git_fetch_init_options(pointer, UInt32(GIT_FETCH_OPTIONS_VERSION))
+
+	var options = pointer.move()
+
+	pointer.dealloc(1)
+
+	options.callbacks.payload = credentials.toPointer()
+	options.callbacks.credentials = credentialsCallback
+
+	return options
+}
+
+private func cloneOptions(bare: Bool = false, localClone: Bool = false, fetchOptions: git_fetch_options? = nil,
+	checkoutOptions: git_checkout_options? = nil) -> git_clone_options {
+
+	let pointer = UnsafeMutablePointer<git_clone_options>.alloc(1)
+	git_clone_init_options(pointer, UInt32(GIT_CLONE_OPTIONS_VERSION))
+
+	var options = pointer.move()
+
+	pointer.dealloc(1)
+
+	options.bare = bare ? 1 : 0
+
+	if localClone {
+		options.local = GIT_CLONE_NO_LOCAL
+	}
+
+	if let checkoutOptions = checkoutOptions {
+		options.checkout_opts = checkoutOptions
+	}
+
+	if let fetchOptions = fetchOptions {
+		options.fetch_opts = fetchOptions
+	}
+
+	return options
+}
+
 /// A git repository.
 final public class Repository {
-	
+
 	// MARK: - Creating Repositories
 	
 	/// Load the repository at the given URL.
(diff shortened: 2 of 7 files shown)


tombooth:
I have found the issue with the test runs. `xctool` runs the tests as logic only whereas `xcodebuild` and XCode run them as application tests. When tests are running logic only a lot of the subsystems that would be available on the simulator are not there, for example keychain and networking.

There are two choices:
  - Add a fake application target to the project so the test can force `xctool` to run in application test mode
  - Migrate the CI scripts back to `xcodebuild`

I would quite like some input as to what would be preferred for SwiftGit2.

mdiep:
I think moving back to `xcodebuild` is the best option. If we want, `xcpretty` does a nice job of cleaning up the output.

Reviewing this is still on my list. I'm hoping to get to it today or tomorrow.

SwiftGit2/SwiftGit2.m
```diff
@@ -34,3 +34,21 @@ git_checkout_options SG2CheckoutOptions(SG2CheckoutProgressBlock progress) {
...
+	return result;
+}
+
+git_fetch_options SG2FetchOptions() {
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wmissing-field-initializers"
+	git_fetch_options result = GIT_FETCH_OPTIONS_INIT;
+#pragma clang diagnostic pop
+
+	return result;
```
mdiep: It would be nice if these could follow the example of `SG2CheckoutOptions` and do all the configuring inside this function. Ideally, `cloneFromURL` wouldn't need to do any manual configuration.

SwiftGit2/Credentials.swift
```diff
@@ -0,0 +1,78 @@
...
+//
+//  Created by Tom Booth on 29/02/2016.
+//  Copyright © 2016 GitHub, Inc. All rights reserved.
+//
+
+import Foundation
+import Result
+
+/// A Swift mapping of git_credtype_t
+public struct CredentialType : OptionSetType {
```
mdiep: I'm struggling to fully understand the credential API from libgit2 (`git_cred` in particular), but I don't like how closely the credentials stuff here mirrors the C API. I'd like to aim for a more native Swift API and let SwiftGit2 handle the translation to the C API.

So, for example, SwiftGit2 might do something like this:

```swift
public enum Credentials {
    case Plaintext(username: String, password: String)
    case SSHMemory(username: String, publicKey: String, privateKey: String, passphrase: String)
    ....

    private var credtype: git_credtype_t { ... }
}
```

But I'm not entirely sure how workable that is because I don't fully understand the libgit2 API. I'd love for you to either pursue a more native Swift API or educate me about why that wouldn't work.

SwiftGit2/Repository.swift
```diff
@@ -32,6 +67,47 @@ final public class Repository {
...
 		return Result.Success(repository)
 	}
+
+	/// Clone the repository from a given URL.
+	///
+	/// remoteURL - The URL of the remote repository
+	/// localURL - The URL to clone the remote repository into
+	///
+	/// Returns a `Result` with a `Repository` or an error.
+	class public func cloneFromURL(remoteURL: NSURL, toWorkingDirectory: NSURL, localClone: Bool = false, bare: Bool = false, checkout: Bool = true,
```
mdiep: `toWorkingDirectory` could probably be called `toURL`

mdiep: Could local be inferred from the two `NSURL`s? (Whether they're both file URLs?)

SwiftGit2Tests/RepositorySpec.swift
```diff
@@ -29,6 +29,99 @@ class RepositorySpec: QuickSpec {
...
+						expect(remote.URL).to(equal(remoteRepoURL?.absoluteString))
+					}
+				}
+			}
+
+			let privateRepo = NSProcessInfo.processInfo().environment["SG2TestPrivateRepo"]
+			let username = NSProcessInfo.processInfo().environment["SG2TestUsername"]
+			let publicKey = NSProcessInfo.processInfo().environment["SG2TestPublicKey"]
+			let privateKey = NSProcessInfo.processInfo().environment["SG2TestPrivateKey"]
+			let passphrase = NSProcessInfo.processInfo().environment["SG2TestPassphrase"]
```
mdiep: Can you extract the `environment` to a local variable since we're using it several times? IIRC, it's quite expensive to get, which could affect the runtime of the unit tests.

mdiep:
Thanks for opening a pull request! I left some initial thoughts. Let me know when it's ready for another look. :sparkles:

SwiftGit2/SwiftGit2.m
tombooth: Out of curiousity, why the preference for moving logic into Objective-C rather than Swift?

I initially did follow a similar pattern as CheckoutOptions but I just ended having to pull more and more into Objective-C land, so I changes to moving it into Swift now 2 can handle C function pointers.

tombooth:
FYI, not abandoned this just been busy at the moment. Hope to get back to it soon!

SwiftGit2Tests/RepositorySpec.swift
tombooth: Done.

SwiftGit2/Repository.swift
tombooth: This controls whether it copies or fetches the object database. I'm not sure whether we'd want to assume that is always the case when both the URLs are local, happy to infer if you think it would be better though.

SwiftGit2/SwiftGit2.m
mdiep: It's not that I want them to be in Objective-C. It's that I want them to (1) be consistent and (2) encapsulate the manipulation of the git fields. #60 moves `SG2CheckoutOptions` into Swift, so this PR can do the same.

SwiftGit2/Credentials.swift
tombooth: I've simplified the credentials handling following this pattern, thoughts?

SwiftGit2/SwiftGit2.m
tombooth: That is fair, I will follow suit.

tombooth:
I still need to address the cibuild script so that it doesn't fail on iOS and tidy up the commits, but I am mainly interested in thoughts on the simplified credentials API.

SwiftGit2/Credentials.swift
```diff
@@ -0,0 +1,50 @@
...
+
+private class Wrapper<T> {
+	let value: T
+
+	init(_ value: T) {
+		self.value = value
+	}
+}
+
+public enum Credential {
```
mdiep: This enum should probably be `Credentials`, plural.

SwiftGit2/Repository.swift
```diff
@@ -11,9 +11,46 @@ import Result
 
 public typealias CheckoutProgressBlock = SG2CheckoutProgressBlock
 
+private func SG2FetchOptions(credentials: Credential = .Default()) -> git_fetch_options {
```
mdiep: I'd remove the default here. It's always specified, and that's probably for the best.

SwiftGit2/Repository.swift
```diff
@@ -11,9 +11,46 @@ import Result
...
+	let pointer = UnsafeMutablePointer<git_fetch_options>.alloc(1)
```
mdiep: This `alloc` needs a matching `dealloc`.

SwiftGit2/Repository.swift
```diff
@@ -32,6 +69,33 @@ final public class Repository {
 		let repository = Repository(pointer)
 		return Result.Success(repository)
 	}
+
+	/// Clone the repository from a given URL.
+	///
+	/// remoteURL - The URL of the remote repository
+	/// localURL - The URL to clone the remote repository into
+	///
```
mdiep: Can you add documentation for the other parameters?

SwiftGit2Tests/RepositorySpec.swift
```diff
@@ -29,6 +29,91 @@ class RepositorySpec: QuickSpec {
 				)))
 			}
 		}
+
+		describe("Reposityory.Type.clone()") {
```
mdiep: Typo here. :relaxed:

mdiep:
That looks a lot better. :clap: 

Sorry it took me a few days to review.

tombooth:
Right, how's that? :tada:

mdiep:
I'm on vacation with the family, so it may take some time for me to review this. But it's on my list, and I will get to it!

tombooth:
Thanks a lot for letting me know, enjoy your holiday!

SwiftGit2/Repository.swift
```diff
@@ -32,6 +73,37 @@ final public class Repository {
...
+	/// credentials      - Credentials to be used when connecting to the remote.
+	/// checkoutStrategy - The checkout strategy to use, if being checked out.
+	/// checkoutProgress - A block that's called with the progress of the checkout.
+	///
+	/// Returns a `Result` with a `Repository` or an error.
+	class public func cloneFromURL(remoteURL: NSURL, toURL: NSURL, localClone: Bool = false, bare: Bool = false,
+		credentials: Credentials = .Default(), checkoutStrategy: CheckoutStrategy = .Safe, checkoutProgress: CheckoutProgressBlock? = nil) -> Result<Repository, NSError> {
+			var checkoutOptions = SG2CheckoutOptions(checkoutProgress)
+
+			checkoutOptions.checkout_strategy = checkoutStrategy.git_checkout_strategy.rawValue
```
mdiep: Could we pass `checkoutStrategy` into `SG2CheckoutOptions()`? I'd like to encapsulate the configuration of the libgit2 structs.

mdiep:
This is looking really good! I just had one note, which shouldn't be much trouble to fix up.

The memory management makes me nervous, but I think everything looks right.

tombooth:
Hi, I've got a little side tracked trying to get everything working on XCode7.3 (expect a PR for that soon). Once that is done I'll address that last point.

Is the memory management concern around the git_cred created and not free'd in SwiftGit2? If so, it seems it is the job of the transport (see https://github.com/libgit2/libgit2/blob/v0.23.4/src/transports/ssh.c and https://github.com/libgit2/libgit2/blob/v0.23.4/src/transports/http.c) to free up the git_cred once it is done with it. The have introduced a `git_cred_free` function in libgit2 0.24.0 but with the note that it should only be used by transports. I can understand why this might make you nervous, I guess we just have to have faith in libgit2.

mdiep:
> I've got a little side tracked trying to get everything working on XCode7.3 (expect a PR for that soon).

✨ 

> Is the memory management concern around the git_cred created and not free'd in SwiftGit2?

No, it's just a lot of work to figure out whether we're handling all the memory management correctly with the bridging and wrapping—harder than it would be in C or Objective-C. But it all looks right.

mdiep:
Thanks for all your work on this! I really appreciate your willingness to figure it out and to address feedback. Sorry that my reviews have been slow. 😞 I'd like them to be faster, but other projects (Carthage in particular) have been higher priorities.

Since you've been making since great contributions, I've invited you to join the org. That means you can push branches to this repo directly instead of using your fork, although we do like all contributors to open PRs for changes.

Thanks again! I'm excited to merge this.

tombooth:
Thanks @mdiep, don't worry about the reviews I can totally understand. Thanks for the invite!

PR #60, opened 2016-03-14
Remove the implementation detail leak, implemented in swift

drkibitz:
Hello thank you for this great work!

This is my attempt at move "leaky details" in the public header to Swift.
My motivations are writing a pure swift package via Swift Package Manager.
You will also notice that the git2.h header no longer needs to be public either.

I also have a followup the last .m item (the constructor), to a central class, calling git_libgit2_init on init and shutdown on deinit, like this: https://github.com/drkibitz/swift-Git2/blob/master/Sources/Git2/Git2.swift

### DIFF
--- a/SwiftGit2/Repository.swift
+++ b/SwiftGit2/Repository.swift
@@ -23,7 +23,7 @@ private func checkoutProgressCallback(path: UnsafePointer<Int8>, completed_steps
 			block = buffer.move()
 			buffer.dealloc(1)
 		}
-		block((path == nil ? nil : String(path)), completed_steps, total_steps);
+		block(String.fromCString(path), completed_steps, total_steps);
 	}
 }
 
--- a/SwiftGit2/Repository.swift
+++ b/SwiftGit2/Repository.swift
@@ -9,7 +9,44 @@
 import Foundation
 import Result
 
-public typealias CheckoutProgressBlock = SG2CheckoutProgressBlock
+public typealias CheckoutProgressBlock = (String?, Int, Int) -> Void
+
+/// Helper function used as the libgit2 progress callback in git_checkout_options.
+/// This is a function with a type signature of git_checkout_progress_cb.
+private func checkoutProgressCallback(path: UnsafePointer<Int8>, completed_steps: Int, total_steps: Int, payload: UnsafeMutablePointer<Void>) -> Void {
+	if (payload != nil) {
+		let buffer = UnsafeMutablePointer<CheckoutProgressBlock>(payload)
+		let block: CheckoutProgressBlock
+		if completed_steps < total_steps {
+			block = buffer.memory
+		} else {
+			block = buffer.move()
+			buffer.dealloc(1)
+		}
+		block((path == nil ? nil : String(path)), completed_steps, total_steps);
+	}
+}
+
+/// Helper function for initializing libgit2 got_checkout_options.
+///
+/// :param: progress A block that's called with the progress of the checkout.
+/// :returns: Returns a git_checkout_options struct with the progress members set.
+private func checkoutOptions(progress: CheckoutProgressBlock? = nil) -> git_checkout_options {
+	// Do this because GIT_CHECKOUT_OPTIONS_INIT is unavailable in swift
+	let pointer = UnsafeMutablePointer<git_checkout_options>.alloc(1)
+	git_checkout_init_options(pointer, UInt32(GIT_CHECKOUT_OPTIONS_VERSION))
+	var options = pointer.move()
+	pointer.dealloc(1)
+
+	if progress != nil {
+		options.progress_cb = checkoutProgressCallback
+		let blockPointer = UnsafeMutablePointer<CheckoutProgressBlock>.alloc(1)
+		blockPointer.initialize(progress!)
+		options.progress_payload = UnsafeMutablePointer<Void>(blockPointer)
+	}
+
+	return options
+}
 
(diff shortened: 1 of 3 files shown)


SwiftGit2.xcodeproj/project.pbxproj
```diff
@@ -74,7 +74,7 @@
 		621E66A71C72958800A0F352 /* Pointers.swift in Sources */ = {isa = PBXBuildFile; fileRef = BE7A753E1A4A2BCC002DA7E3 /* Pointers.swift */; };
 		621E66A81C72958800A0F352 /* Errors.swift in Sources */ = {isa = PBXBuildFile; fileRef = DA5914751A94579000AED74C /* Errors.swift */; };
 		621E66A91C72958800A0F352 /* SwiftGit2.m in Sources */ = {isa = PBXBuildFile; fileRef = BE14AA4F1A1974010015B439 /* SwiftGit2.m */; };
-		621E66AF1C72958800A0F352 /* git2.h in Headers */ = {isa = PBXBuildFile; fileRef = BEB31F9B1A0E581400F525B9 /* git2.h */; settings = {ATTRIBUTES = (Public, ); }; };
```
mdiep: Could you do this is a separate PR? I'm uncertain about this part, and I don't want it to hold up the rest of the changes.

SwiftGit2/Repository.swift
```diff
@@ -9,7 +9,32 @@
...
+func SG2CheckoutProgressCallback(path: UnsafePointer<Int8>, completed_steps: Int, total_steps: Int, payload: UnsafeMutablePointer<Void>) -> Void {
+	if (payload != nil) {
+		let buffer = UnsafeMutablePointer<CheckoutProgressBlock>(payload)
+		let block: CheckoutProgressBlock = buffer.memory as CheckoutProgressBlock
+		block((path == nil ? nil : String(path)), completed_steps, total_steps);
+	}
+}
+
+func SG2CheckoutOptions(progress: CheckoutProgressBlock? = nil) -> git_checkout_options {
+
```
mdiep: Could you remove this empty line?

SwiftGit2/Repository.swift
```diff
@@ -9,7 +9,32 @@
 import Foundation
 import Result
 
-public typealias CheckoutProgressBlock = SG2CheckoutProgressBlock
+public typealias CheckoutProgressBlock = (String?, Int, Int) -> Void
+
+func SG2CheckoutProgressCallback(path: UnsafePointer<Int8>, completed_steps: Int, total_steps: Int, payload: UnsafeMutablePointer<Void>) -> Void {
+	if (payload != nil) {
+		let buffer = UnsafeMutablePointer<CheckoutProgressBlock>(payload)
+		let block: CheckoutProgressBlock = buffer.memory as CheckoutProgressBlock
```
mdiep: We should only need to specify the type once on this line

SwiftGit2/Repository.swift
```diff
@@ -9,7 +9,32 @@
...
+func SG2CheckoutOptions(progress: CheckoutProgressBlock? = nil) -> git_checkout_options {
+
+	// Do this because GIT_CHECKOUT_OPTIONS_INIT is unavailable in swift
+	let pointer = UnsafeMutablePointer<git_checkout_options>.alloc(1)
+	git_checkout_init_options(pointer, UInt32(GIT_CHECKOUT_OPTIONS_VERSION))
+
+	var result: git_checkout_options = pointer.memory
+	if (progress != nil) {
+		result.progress_cb = SG2CheckoutProgressCallback
+		let aProgress = UnsafeMutablePointer<CheckoutProgressBlock>.alloc(1)
```
mdiep: If we call `alloc`, we also need to call `dealloc`.

mdiep:
Thanks for the pull request! I'm really excited to move this code to Swift. When I originally wrote the code, that wasn't possible.

drkibitz:
@mdiep thanks for the comments, new commits here with :cake:

drkibitz:
Sorry got a little trigger happy ;)

SwiftGit2/Repository.swift
```diff
@@ -9,7 +9,34 @@
...
 import Result
 
-public typealias CheckoutProgressBlock = SG2CheckoutProgressBlock
+public typealias CheckoutProgressBlock = (String?, Int, Int) -> Void
+
+func SG2CheckoutProgressCallback(path: UnsafePointer<Int8>, completed_steps: Int, total_steps: Int, payload: UnsafeMutablePointer<Void>) -> Void {
+	if (payload != nil) {
+		let buffer = UnsafeMutablePointer<CheckoutProgressBlock>(payload)
+		let block = buffer.move()
+		payload.dealloc(1)
```
mdiep: This doesn't seem right. This block should be called multiple times in a checkout.

drkibitz: Right, was thinking of a it as a completion for some reason. Also something to note is that this doesn't happen in the test.

drkibitz:
Updated, but... I'm not really sure, should I move the alloc for the callback pointer to the checkout method, and dealloc there too?

drkibitz:
@mdiep Pretty sure I just I answered my own question... just check for completion in the progress callback and dealloc then. Also, I'm going to squash these commits, and push it back up since this PR is getting a git ugly.

mdiep:
Looks good—thanks!

(Sorry it took me so long to review this!)

PR #67, opened 2016-12-14
Use a modulemap for libgit2

mattrubin:
Instead of including the libgit2 headers publicly in the SwiftGit2 framework and modulemap, wrap the libgit2 headers in their own modulemap and explicitly import the module into the SwiftGit2 source files.


mdiep:
This is great! I struggled to get this working originally. This seems like a much better solution.

mattrubin:
This might break anyone who was relying on the libgit2 API being accessible via SwiftGit2, but this library is pre-1.0 and the Swift 2.3 update already made breaking changes, so hopefully that's not a major concern.

PR #71, opened 2016-12-18
Configure SwiftLint

mattrubin:
The added `.swiftlint.yml` file attempts to codify the Swift style already used in this project. The accompanying source changes are corrections to the handful of lines not already conforming to the conventions of the rest of the code.

I have disabled seven of SwiftLint's default rules to create a configuration that currently shows no warnings or errors for the project. Of the rules currently disabled:
- Some (such as `force_cast`) show errors in many places across the project, and could not be used to lint existing code without large, widespread changes.
- Some (`file_length`, `type_body_length`) show a few warnings, and represent style guidelines that I think it would be good to update the source to follow at some point in the future.
- Some (`force_try`) show warnings for style violations in the tests only, and the rule could be enabled to lint the project if those few lines were annotated with SwiftLint exceptions inline: https://github.com/realm/SwiftLint/blob/master/README.md#disable-rules-in-code

For all of the currently disabled rules, the changes to conform to them would be somewhat invasive, so I've deferred those decisions to future PRs.

-----
There are several possible ways the project can use SwiftLint going forward:
1. Add a build phase which runs SwiftLint on every build, to enforce style rules during development.
2. Add a CI tool like [Hound](https://houndci.com) (or [Danger](https://github.com/ashfurrow/danger-swiftlint)) to lint the code changed in pull requests...
    1. ...with this configuration, to continue enforcing the project's style status quo.
    2. ...with a stricter configuration, to discourage additional style violations which are present in the existing code (e.g. very long function bodies, uses of `try!`). 
3. Add this configuration file to configure manual, optional usage of SwiftLint, but don't integrate SwiftLint into the build process.


mdiep:
This is great. ✨ 

> There are several possible ways the project can use SwiftLint going forward:

My preference would be a build phase, pre-commit hook, or CI tool. 👍

mattrubin:
I've added a separate Xcode target which runs SwiftLint, and then added it to the iOS and macOS framework schemes to run for every build type. (Adding it this way means we don't need to add a separate run script build phase to every build target, and also allows for the creation of a scheme that doesn't lint or the selective disabling of linting without modifying the framework target's build phases.)

I'll look into enabling Hound in a future PR.

PR #78, opened 2016-12-23
Use guard-else instead of if statements when checking git error codes

mattrubin:
The compiler will warn if a `guard` statement falls through without exiting the scope.


mdiep:
This has conflicts that need to be resolved.

PR #83, opened 2017-01-07
Fix the release badge in the README

mattrubin:
The release badge added in #82 was for ReactiveSwift, not SwiftGit2.


mdiep:
Oops. 😆

PR #86, opened 2017-04-24
Improvements for SwiftLint 0.18

mattrubin:
When attempting to build SwiftGit2 via Carthage, SwiftLint caused the build to fail because of new linter rules added after the project was configured to use SwiftLint. This PR changes the SwiftLint build phase to use the `--lenient` flag, which will cause the linter to generate only warnings, not errors.

This PR also...
 - fixes a small lint issue with explicitly returning Void from a function.
 - enables several new linter rules which seem helpful.
 - disables the `identifier_name` rule, which was complaining about enum cases not beginning with lowercase letters. (See #87)
 - configures Travis to use the Xcode 8.3 image, which has SwiftLint 0.18.1 installed.

Fixes #85.

PR #89, opened 2017-04-25
Add ability to get all commits from a branch

myste1tainn:
I'm pretty new to libgit2 and SwiftGit2, I don't know If the code I added is correct or no, please review.
The purpose of this enhance is ability to list commits. At first I think SwiftGit2 should already supported that, I've tried to find some documentation on that, but came up with nothing. So I added the codes.

I'm new to pull request also, if I've done anything wrong, do tell, please.


SwiftGit2.xcodeproj/project.pbxproj
```diff
@@ -729,7 +729,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "if which swiftlint >/dev/null; then\n    swiftlint --lenient\nelse\n    echo \"warning: SwiftLint is not installed\"\nfi";
+			shellScript = "if which swiftlint >/dev/null; then\n    swiftlint\nelse\n    echo \"warning: SwiftLint is not installed\"\nfi";
```
mattrubin: This removes a flag that was added in #86. I assume this was accidental? Please restore the `--lenient` flag.

SwiftGit2/References.swift
```diff
@@ -64,6 +64,7 @@ extension Reference: Hashable {
 
 /// A git branch.
 public struct Branch: ReferenceType {
+
```
mattrubin: This extra newline is unnecessary.

SwiftGit2/References.swift
```diff
@@ -119,6 +120,27 @@ public struct Branch: ReferenceType {
...
+		var commits: [Commit] = []
+		var walker: OpaquePointer? = nil
+		var unsafeCommit: OpaquePointer? = nil
+		var oid = self.oid.oid
+		git_revwalk_new(&walker, repo.pointer)
+		git_revwalk_sorting(walker, GIT_SORT_TOPOLOGICAL.rawValue)
+		git_revwalk_push(walker, &oid)
+		while git_revwalk_next(&oid, walker) == GIT_OK.rawValue {
+			guard
+				git_commit_lookup(&unsafeCommit, repo.pointer, &oid) == GIT_OK.rawValue,
```
mattrubin: This can be on the same line as the `guard`.

SwiftGit2/References.swift
```diff
@@ -119,6 +120,27 @@ public struct Branch: ReferenceType {
 		}
 		commit = PointerTo<Commit>(oid)
 	}
+	
+	public func allCommits(in repo: Repository) -> [Commit] {
```
mattrubin: Please add a comment explaining the return value of this method. For instance, are the returned commits sorted from newest to oldest?

SwiftGit2/References.swift
```diff
@@ -119,6 +120,27 @@ public struct Branch: ReferenceType {
...
+		var unsafeCommit: OpaquePointer? = nil
+		var oid = self.oid.oid
+		git_revwalk_new(&walker, repo.pointer)
+		git_revwalk_sorting(walker, GIT_SORT_TOPOLOGICAL.rawValue)
+		git_revwalk_push(walker, &oid)
+		while git_revwalk_next(&oid, walker) == GIT_OK.rawValue {
+			guard
+				git_commit_lookup(&unsafeCommit, repo.pointer, &oid) == GIT_OK.rawValue,
+				let commit = unsafeCommit else {
+					return commits
```
mattrubin: If `git_commit_lookup` fails to produce the requested commit, should the function return the commits, or should it return an error?

SwiftGit2/References.swift
```diff
@@ -119,6 +120,27 @@ public struct Branch: ReferenceType {
...
+			}
+			commits += [Commit(commit)]
+			git_commit_free(unsafeCommit)
+		}
+		git_revwalk_free(walker)
```
mattrubin: If the function returns on line 136, this `git_revwalk_free` will never be called and the object will leak. Perhaps move this to a `defer { ... }` statement just after the `git_revwalk_new`, to ensure it is called for every return path.

SwiftGit2.xcodeproj/project.pbxproj
myste1tainn: This is giving me the error like unknown flag when I tried to compiled the code, not sure what happened there. I've tried googling also, nothing come up. Any suggestions? My SwiftLint version is wrong?

myste1tainn: Like this
screen shot 2017-04-27 at 07 31 39

And this is my version
screen shot 2017-04-27 at 07 34 16

myste1tainn:
@mattrubin I think I fixed all the problem you've mentioned, how do I resubmit those code?, lol. Just create a new pull request?, I guess.

myste1tainn:
Just noticed that the changes I push in my branches, comes up here in the thread, cool!

mdiep:
It's really great that you're opening a PR! It's always exciting to have new contributors.

If you feel overwhelmed or lost at any point, just reach out. We'll help guide you through. ✨

SwiftGit2/Repository.swift
```diff
@@ -509,4 +509,35 @@ final public class Repository {
 	                     progress: CheckoutProgressBlock? = nil) -> Result<(), NSError> {
 		return setHEAD(reference).flatMap { self.checkout(strategy: strategy, progress: progress) }
 	}
+	
+	
+	/// Load all commits in the specified branch in topological & time order descending
+	///
+	/// :param: branch The branch to get all commits from
+	/// :returns: Returns a result with array of branches or the error that occurred
+	public func allCommits(in branch: Branch) -> Result<[Commit], NSError> {
```
myste1tainn: Ok, I'll try that, never have a chances to write an iterator before.

myste1tainn:
@mdiep pushed the suggested change, can you help review them?

SwiftGit2/Repository.swift
```diff
@@ -509,4 +509,23 @@ final public class Repository {
 	                     progress: CheckoutProgressBlock? = nil) -> Result<(), NSError> {
 		return setHEAD(reference).flatMap { self.checkout(strategy: strategy, progress: progress) }
 	}
+	
+	/// Cache for commits(in: branch)
+	///
+	/// Specifically for allowing syntax "while let commit = repo.commits(in: branch).next() {}"
+	public var branchCommitIteratorMap: [Branch: CommitIterator] = [:]
```
mdiep: This has to go, unfortunately. Multiple readers should be able to read the same branch simultaneously. This would seriously complicate that.

But… This should do what you want:

```swift
for commit in repo.commits(in: branch) {}
```

SwiftGit2Tests/RepositorySpec.swift
```diff
@@ -600,6 +600,21 @@ class RepositorySpec: QuickSpec {
...
+		fdescribe("Repository.allCommits(in:)") {
+			it("should return all (9) commits") {
+				let repo = Fixtures.simpleRepository
+				let branches = repo.localBranches().value!
+                var count = 0
+                let expected = 9
+				for branch in branches {
+                    while repo.commits(in: branch).next() != nil {
+                        count += 1
+                    }
```
mdiep: Some indentation is messed up here.

SwiftGit2Tests/RepositorySpec.swift
```diff
@@ -600,6 +600,21 @@ class RepositorySpec: QuickSpec {
 				expect(repo.HEAD().value?.longName).to(equal("HEAD"))
 			}
 		}
+		
+		fdescribe("Repository.allCommits(in:)") {
```
mdiep: The `f` needs to come off the front of this `describe`.

SwiftGit2/CommitIterator.swift
```diff
@@ -0,0 +1,67 @@
...
+
+public class CommitIterator: IteratorProtocol {
+    public typealias Element = Result<Commit, NSError>
+    var repo: Repository
+    var branch: Branch
+    var revisionWalker: OpaquePointer? = nil
+    var oid: git_oid
+    private var unsafeCommit: OpaquePointer? = nil
+    
+    public init(repo: Repository, branch: Branch) {
```
mdiep: Let's make this `internal`. People should create iterator using `Repository.commits(in: Branch)`. We'll also want to create different types of iterators, so this is unnecessarily restrictive.

myste1tainn: Can you elaborate more on "We'll also want to create different types of iterators"?, Did you mean like to return generic iterator class instead? e.g. commits(in:) -> Iterator<Commit>, although the actual Instance returned will be of CommitIterator.

mdiep: Oh, there are lots of reasons to iterate over commits. Most often you'll iterate over the difference between 2 commitish objects (commit, tag, branch). This used to find:

1. New commits in a branch/PR
2. The merge base
3. The ahead/behind count of a branch

"Rev walking", as libgit2 calls it, is a core conscript for git.

myste1tainn: Ok, let's see if I understand you right. Because the code returns only Iterator of only "Commit" types object so it's too restrictive, while Commit, Tag, Branch are & can be equally viewed as the "Commit".

So, what's the function should return is of somewhat iterator of super class of those or some protocols those objects has in common, correct?

mdiep: No, I think it should always return commits. I was just saying that you can create iterates of commits for different things—not just branches.

Probably the important bit is that there should just be one way to create an iterator for a specific thing. This initializer doesn't need to be public when there's another public way to create one.

SwiftGit2/CommitIterator.swift
```diff
@@ -0,0 +1,67 @@
...
+    
+    private func result(withName name: String, from result: Int32) -> (stop: Bool, error: NSError?) {
+        guard result == GIT_OK.rawValue else {
+            if result == GIT_ITEROVER.rawValue {
+                return (stop: true, error: nil)
+            } else {
+                return (stop: false, error: NSError(gitError: result, pointOfFailure: name))
+            }
+        }
+        return (stop: false, error: nil)
```
mdiep: I think this would be clearer with a private enum.

```swift
public class CommitIterator: IteratorProtocol {
    …

    private enum Next {
        case over
        case ok
        case error(NSError)

        init(_ result: Int32, name: String) {
            switch result {
            case GIT_ITEROVER.rawValue:
                return .over
            case GIT_OK.rawValue:
                return .ok
            default:
                return .error(NSError(gitError: result, pointOfFailure: name))
            }
        }
    }
}
```

SwiftGit2/CommitIterator.swift
```diff
@@ -0,0 +1,67 @@
...
+
+import Result
+import libgit2
+
+public class CommitIterator: IteratorProtocol {
+    public typealias Element = Result<Commit, NSError>
+    var repo: Repository
+    var branch: Branch
+    var revisionWalker: OpaquePointer? = nil
+    var oid: git_oid
```
mdiep: I don't think `oid` should be a property. It can be replaced with a local variable inside `next`. It's only set and then immediately read.

SwiftGit2/CommitIterator.swift
```diff
@@ -0,0 +1,67 @@
...
+// Copyright (c) 2017 GitHub, Inc. All rights reserved.
+//
+
+import Result
+import libgit2
+
+public class CommitIterator: IteratorProtocol {
+    public typealias Element = Result<Commit, NSError>
+    var repo: Repository
+    var branch: Branch
```
mdiep: The branch doesn't need to be saved. It's only used inside `init`.

SwiftGit2/CommitIterator.swift
```diff
@@ -0,0 +1,67 @@
...
+// Created by Arnon Keereena on 4/28/17.
+// Copyright (c) 2017 GitHub, Inc. All rights reserved.
+//
+
+import Result
+import libgit2
+
+public class CommitIterator: IteratorProtocol {
+    public typealias Element = Result<Commit, NSError>
+    var repo: Repository
```
mdiep: This should be a `let`. Changing the repository would do Bad Things™.

SwiftGit2/CommitIterator.swift
```diff
@@ -0,0 +1,67 @@
...
+    var branch: Branch
+    var revisionWalker: OpaquePointer? = nil
+    var oid: git_oid
+    private var unsafeCommit: OpaquePointer? = nil
```
mdiep: Same is true for `unsafeCommit`—it doesn't need to be a property.

SwiftGit2/CommitIterator.swift
```diff
@@ -0,0 +1,67 @@
...
+        if lookupResult.stop {
+            return nil
+        } else if let error = lookupResult.error {
+            return Result.failure(error)
+        }
+        guard let commit = unsafeCommit else {
+            return nil
+        }
+        git_commit_free(unsafeCommit)
+        return Result.success(Commit(commit))
```
mdiep: You need to create `Commit(commit)` before `free`ing `unsafeCommit`.

SwiftGit2/CommitIterator.swift
```diff
@@ -0,0 +1,67 @@
...
+        let revwalkResult = result(withName: "git_revwalk_next", from: revwalkGitResult)
+        if revwalkResult.stop {
+            return nil
+        } else if let error = revwalkResult.error {
+            return Result.failure(error)
+        }
+        let lookupGitResult = git_commit_lookup(&self.unsafeCommit, self.repo.pointer, &self.oid)
+        let lookupResult = result(withName: "git_commit_lookup", from: lookupGitResult)
+        if lookupResult.stop {
+            return nil
```
mdiep: This should never result in a `stop`, so this doesn't need to use `result(withName:from:)`.

SwiftGit2/CommitIterator.swift
myste1tainn: Hmm? OID does not have to be initially set from branch's OID, right?

mdiep: It does, but you could pass that OID to `setupRevisionWalker`. It doesn't need to be saved on a property.

myste1tainn: Ok, I see now, after reading the header docs of git_revwalk_push, git_revwalk_next and git_commit_lookup. So the OID needed initially to create the revwalk, then the next git_oid used in git_revwalk_next does not need to be the old, because its git pointed to a new one any way. Then from the oid that is populated. It is used to lookup for the actual commit in git_commit_lookup. After that there's no need for the oid anymore. Thank you for pointing that out. I still understand very less about these things.

SwiftGit2/CommitIterator.swift
```diff
@@ -0,0 +1,75 @@
...
+		git_revwalk_push(revisionWalker, &oid)
+	}
+
+	private func next(withName name: String, from result: Int32) -> Next {
+		if result == GIT_OK.rawValue || result == GIT_ITEROVER.rawValue {
+			return Next(result, name: name)
+		} else {
+			return Next.error(NSError(gitError: result, pointOfFailure: name))
+		}
+	}
```
mdiep: This does the same thing as `Next(result, name: name)`—so this method can be removed.

SwiftGit2/CommitIterator.swift
```diff
@@ -0,0 +1,75 @@
...
+		let nextResult = next(withName: "git_revwalk_next", from: revwalkGitResult)
+		if case let .error(error) = nextResult {
+			return Result.failure(error)
+		} else if case .over = nextResult {
+			return nil
+		}
+		let lookupGitResult = git_commit_lookup(&unsafeCommit, repo.pointer, &oid)
+		guard lookupGitResult == GIT_OK.rawValue,
+		      let unwrapCommit = unsafeCommit else {
+			return nil
```
mdiep: This should return an error. We didn't run out of commits; we failed to load one for some reason.

SwiftGit2/Repository.swift
```diff
@@ -509,4 +509,23 @@ final public class Repository {
...
+	/// Load all commits in the specified branch in topological & time order descending
+	///
+	/// :param: branch The branch to get all commits from
+	/// :returns: Returns a result with array of branches or the error that occurred
+	public func commits(in branch: Branch) -> CommitIterator {
+		if let iterator = branchCommitIteratorMap[branch] {
+			return iterator
+		} else {
+			let iterator = CommitIterator(repo: self, root: branch.oid.oid)
+			branchCommitIteratorMap[branch] = iterator
```
mdiep: This map still needs to be removed as I explained earlier.

SwiftGit2Tests/RepositorySpec.swift
```diff
@@ -600,6 +600,33 @@ class RepositorySpec: QuickSpec {
...
+					"Alphabetize branches\n",
+					"List new branches\n",
+					"List branches in README\n",
+					"Create a README\n",
+					"List branches in README\n",
+					"Create a README\n"
+				]
+				var commitMessages: [String] = []
+				for branch in branches {
+					while let commit = repo.commits(in: branch).next() {
```
mdiep: We'll want this to be `for commit in repo.commits(in: branch) {` when the map above goes away.

SwiftGit2Tests/RepositorySpec.swift
```diff
@@ -520,88 +520,115 @@ class RepositorySpec: QuickSpec {
...
 				expect(repo.checkout(branch, strategy: CheckoutStrategy.None)).to(haveSucceeded())
 				expect(repo.HEAD().value?.shortName).to(equal(branch.name))
-
+				
 				expect(repo.checkout(oid, strategy: CheckoutStrategy.None)).to(haveSucceeded())
 				expect(repo.HEAD().value?.longName).to(equal("HEAD"))
 			}
 		}
+		
+		fdescribe("Repository.allCommits(in:)") {
```
mdiep: The `f` still needs to come off the front of this so that all the tests run.

SwiftGit2/CommitIterator.swift
```diff
@@ -0,0 +1,140 @@
...
+		}
+		let lookupGitResult = git_commit_lookup(&unsafeCommit, repo.pointer, &oid)
+		guard lookupGitResult == GIT_OK.rawValue,
+		      let unwrapCommit = unsafeCommit else {
+			return Result.failure(NSError(gitError: lookupGitResult, pointOfFailure: "git_commit_lookup"))
+		}
+		let result: Element = Result.success(Commit(unwrapCommit))
+		git_commit_free(unsafeCommit)
+		return result
```
mdiep: Would you mind changing this to `switch` over `nextResult` since you need to make a change to the tests?

```swift
switch nextResult {
case let .error(error):
    return .failure(error)
case .over:
    return nil
case .ok:
    let lookupGitResult = git_commit_lookup(&unsafeCommit, repo.pointer, &oid)
    …
    return result
}
```

mdiep:
This is great! Thanks for all your work on this!

I'll merge as soon as I get the green light from CI.

myste1tainn:
Hooray!~ Thank you.

mdiep:
🎉

PR #98, opened 2017-08-05
Create new repository

jakeva:
Great project! I'm using it in one of my own and adding things I need and couldn't find already. Let me know what you think!


SwiftGit2/Repository.swift
```diff
@@ -99,6 +99,25 @@ final public class Repository {
 
 	// MARK: - Creating Repositories
 
+	/// Create a new repository at the given URL.
+	///
+	/// URL - The URL of the repository.
+	///
+	/// Returns a `Result` with a `Repository` or an error.
+	class public func createAt(_ url: URL) -> Result<Repository, NSError> {
```
mdiep: I think this should be `create(at url: URL)` to match the Swift 3 style guide.

PR #99, opened 2017-08-08
Develop

jakeva:
I agree with your thoughts on my other PR. I also added an agent option for auth. Thoughts there? The test I wrote only tests successful git creation, couldn't think of what else to test but I'd be happy to add to it. Also… sorry for some of the vulgar commit message content, I wasn't planning on contributing.


mdiep:
I'd like to handle the 2 sets of changes separately. That's easier to review and makes the process smoother overall. Could you add your test to #98 and create a new PR (or repurpose this one) for the agent changes?

jakeva:
yeah sure. sorry, like i said i wasn’t planning on contributing but i’m going to need to add more still. might as well make it available and do it right.

mdiep:
Yeah, let's just try to keep it to one addition per PR. That way changes stay manageable and independent. ☺️

PR #101, opened 2017-08-09
add ssh agent authentication

jakeva:
totally agree with how you'd like to proceed, thanks for the feedback.

### DIFF
--- a/SwiftGit2/Credentials.swift
+++ b/SwiftGit2/Credentials.swift
@@ -18,7 +18,7 @@ private class Wrapper<T> {
 
 public enum Credentials {
 	case `default`
-	case agent
+	case sshAgent
 	case plaintext(username: String, password: String)
 	case sshMemory(username: String, publicKey: String, privateKey: String, passphrase: String)
 
@@ -39,8 +39,7 @@ internal func credentialsCallback(
 	url: UnsafePointer<CChar>?,
 	username: UnsafePointer<CChar>?,
 	_: UInt32,
-	payload: UnsafeMutableRawPointer?
-	) -> Int32 {
+	payload: UnsafeMutableRawPointer? ) -> Int32 {
 
 	let result: Int32
 
@@ -50,7 +49,7 @@ internal func credentialsCallback(
 	switch Credentials.fromPointer(payload!) {
 	case .default:
 		result = git_cred_default_new(cred)
-	case .agent:
+	case .sshAgent:
 		result = git_cred_ssh_key_from_agent(cred, name!)
 	case .plaintext(let username, let password):
 		result = git_cred_userpass_plaintext_new(cred, username, password)
--- a/SwiftGit2/Credentials.swift
+++ b/SwiftGit2/Credentials.swift
@@ -16,15 +16,6 @@ private class Wrapper<T> {
 	}
 }
 
-/// Convert libgit2 string to Swift string
-///
-/// - parameter cStr: C string pointer
-///
-/// - returns: Swift string
-func git_string_converter(_ cStr: UnsafePointer<CChar>) -> String {
-	return String(cString: cStr)
-}
-
 public enum Credentials {
 	case `default`
 	case agent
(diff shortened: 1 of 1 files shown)


SwiftGit2/Credentials.swift
```diff
@@ -33,13 +43,24 @@ public enum Credentials {
...
+                                  _: UInt32, payload: UnsafeMutableRawPointer?) -> Int32 {
 	let result: Int32
 
+	// Find username_from_url
+	let name: String?
+	if username == nil {
+		name = nil
+	} else {
+		name = git_string_converter(username!)
+	}
```
mdiep: This could be:

```swift
let name = username.map(String.init(cString:))
```

SwiftGit2/Credentials.swift
```diff
@@ -16,8 +16,18 @@ private class Wrapper<T> {
...
 }
 
+/// Convert libgit2 string to Swift string
+///
+/// - parameter cStr: C string pointer
+///
+/// - returns: Swift string
+func git_string_converter(_ cStr: UnsafePointer<CChar>) -> String {
+	return String(cString: cStr)
+}
```
mdiep: This is just wrapping the Swift `String.init`, so it's not needed.

SwiftGit2/Credentials.swift
```diff
@@ -33,13 +43,24 @@ public enum Credentials {
 /// Handle the request of credentials, passing through to a wrapped block after converting the arguments.
 /// Converts the result to the correct error code required by libgit2 (0 = success, 1 = rejected setting creds,
 /// -1 = error)
-internal func credentialsCallback(cred: UnsafeMutablePointer<UnsafeMutablePointer<git_cred>?>?, _: UnsafePointer<Int8>?,
-                                  _: UnsafePointer<Int8>?, _: UInt32, payload: UnsafeMutableRawPointer?) -> Int32 {
+internal func credentialsCallback(cred: UnsafeMutablePointer<UnsafeMutablePointer<git_cred>?>?,
+                                  url: UnsafePointer<CChar>?, username: UnsafePointer<CChar>?,
+                                  _: UInt32, payload: UnsafeMutableRawPointer?) -> Int32 {
```
mdiep: While we're adjusting these, can we reformat them like so?

```swift
internal func credentialsCallback(
    cred: UnsafeMutablePointer<UnsafeMutablePointer<git_cred>?>?,
    url: UnsafePointer<CChar>?,
    username: UnsafePointer<CChar>?,
    _: UInt32,
    payload: UnsafeMutableRawPointer?
) -> Int32 {
    …
}
```

SwiftGit2/Credentials.swift
```diff
@@ -18,6 +18,7 @@ private class Wrapper<T> {
 
 public enum Credentials {
 	case `default`
+	case agent
```
mdiep: Can we actually name this `sshAgent`?

You could create a different scheme that relied on some other agent, so this is most correct.

SwiftGit2/Credentials.swift
```diff
@@ -33,13 +34,24 @@ public enum Credentials {
...
 /// -1 = error)
-internal func credentialsCallback(cred: UnsafeMutablePointer<UnsafeMutablePointer<git_cred>?>?, _: UnsafePointer<Int8>?,
-                                  _: UnsafePointer<Int8>?, _: UInt32, payload: UnsafeMutableRawPointer?) -> Int32 {
+internal func credentialsCallback(
+	cred: UnsafeMutablePointer<UnsafeMutablePointer<git_cred>?>?,
+	url: UnsafePointer<CChar>?,
+	username: UnsafePointer<CChar>?,
+	_: UInt32,
+	payload: UnsafeMutableRawPointer?
+	) -> Int32 {
```
mdiep: Can you remove the indentation from this line?

mdiep:
A couple more requests. Sorry. 🙈

SwiftGit2/Credentials.swift
jakeva: makes sense

SwiftGit2/Credentials.swift
jakeva: not quite sure i understand what you mean but i'll give it a shot

mdiep:
Thanks for adding this!

PR #102, opened 2017-08-22
Implement git status and diffs

jakeva:
I'm working on implementing repository status and diffs, it isn't complete but I think this is a good starting point. Currently you can get the diff deltas for comparing a given commit in a repository to that commit's parent, or parents if it's a merge commit, or just get what was committed if it's the initial commit. You can also get the current status of a repository from the index to the working directory, or from the HEAD to the index. Let me know what you think!


SwiftGit2/Diffs.swift
```diff
@@ -0,0 +1,59 @@
...
+}
+
+public struct GitDiffDelta {
+	public var status: GitStatus?
+	public var flags: UInt32?
+	public var oldFile: GitDiffFile?
+	public var newFile: GitDiffFile?
+}
+
+public enum GitDiffFlag: Int {
```
mdiep: This looks like it could be an `OptionSet`.

SwiftGit2/Diffs.swift
```diff
@@ -0,0 +1,59 @@
...
+	case binary     = 0
+	case notBinary  = 1
+	case validId    = 2
+	case exists     = 4
+
+	public var value: UInt32 {
```
mdiep: I think you could just declare the `rawValue` to be `UInt32`.

SwiftGit2/Diffs.swift
```diff
@@ -0,0 +1,59 @@
+//
+//  Diffs.swift
+//  SwiftGit2
+//
+//  Created by Jake Van Alstyne on 8/20/17.
+//  Copyright © 2017 GitHub, Inc. All rights reserved.
+//
+
+public struct GitDiffFile {
```
mdiep: I don't think we need to put `Git` in the type names. It's a bit redundant. ☺️

SwiftGit2/Diffs.swift
```diff
@@ -0,0 +1,59 @@
...
+	public var oid: OID
+	public var path: String
+	public var size: Int64
+	public var flags: UInt32
+}
+
+public enum GitStatus: Int {
```
mdiep: I think this could just be declared as `UInt32`.

SwiftGit2/Diffs.swift
jakeva: Brilliant. Didn't know about optionset. Thanks. Partially doing this to get better at swift, definitely appreciate the feedback.

SwiftGit2/Diffs.swift
jakeva: Makes sense. Can't remember why I wrote it this way but I'll clean it up.

SwiftGit2/Repository.swift
```diff
@@ -537,4 +537,297 @@ final public class Repository {
 		let iterator = CommitIterator(repo: self, root: branch.oid.oid)
 		return iterator
 	}
+
+	// MARK: - Diffs
+
+	public func getDiffDeltas(for commit: Commit) -> Result<[GitDiffDelta], NSError> {
```
mdiep: I think we can name this `diff(for commit: Commit)`.

SwiftGit2/Repository.swift
```diff
@@ -537,4 +537,297 @@ final public class Repository {
...
+			if mergeDiff == nil {
+				mergeDiff = unwrapDiffResult
+			} else {
+				let mergeResult = git_diff_merge(mergeDiff, unwrapDiffResult)
+				guard mergeResult == GIT_OK.rawValue else {
+					return Result.failure(NSError(gitError: mergeResult, pointOfFailure: "git_diff_merge"))
+				}
+			}
+		}
+		return self.processDiffDeltas(mergeDiff!)
```
mdiep: This code doesn't look any different than the code in `getDiffDeltasWithOneParent`, so I think that function can be removed. This function can be used instead and the loop will do the right thing.

SwiftGit2/Repository.swift
```diff
@@ -537,4 +537,297 @@ final public class Repository {
...
+			git_diff_free(OpaquePointer(delta))
+		}
+
+		let result = Result<[GitDiffDelta], NSError>.success(returnDict)
+		return result
+	}
+
+	// MARK: - Status
+
+	public func getRepositoryStatus() -> [GitStatusEntry] {
```
mdiep: I think we can call this `status()`.

SwiftGit2/Repository.swift
```diff
@@ -537,4 +537,297 @@ final public class Repository {
...
+		var returnArray = [GitStatusEntry]()
+
+		// Do this because GIT_STATUS_OPTIONS_INIT is unavailable in swift
+		let pointer = UnsafeMutablePointer<git_status_options>.allocate(capacity: 1)
+		git_status_init_options(pointer, UInt32(GIT_STATUS_OPTIONS_VERSION))
+		var options = pointer.move()
+		pointer.deallocate(capacity: 1)
+
+		var status: OpaquePointer? = nil
+		git_status_list_new(&status, self.pointer, &options)
```
mdiep: The return code for this function should be checked.

SwiftGit2/Repository.swift
```diff
@@ -537,4 +537,297 @@ final public class Repository {
...
+				}
+				if let newFile = s?.pointee.index_to_workdir.pointee.new_file {
+					itowNewFile = self.convertDiffFile(newFile)
+				}
+
+				indexToWorkDir = GitDiffDelta(status: itowStatus,
+				                              flags: s?.pointee.index_to_workdir.pointee.flags,
+				                              oldFile: itowOldFile,
+				                              newFile: itowNewFile)
+			}
```
mdiep: The code here is duplicated with the code in the previous `if`. Can you add a function that creates a `GitDiffDelta` from a `git_diff_delta`? Probably adding an `init` to `GitDiffDelta` is the best way to do this.

SwiftGit2/Repository.swift
```diff
@@ -537,4 +537,297 @@ final public class Repository {
...
+			status = GitStatus(rawValue: Int(status!.value & GitStatus.workTreeModified.value))
+		} else if (statusValue & GIT_STATUS_WT_DELETED.rawValue) == GIT_STATUS_WT_DELETED.rawValue {
+			status = GitStatus(rawValue: Int(status!.value & GitStatus.workTreeDeleted.value))
+		} else if (statusValue & GIT_STATUS_WT_RENAMED.rawValue) == GIT_STATUS_WT_RENAMED.rawValue {
+			status = GitStatus(rawValue: Int(status!.value & GitStatus.workTreeRenamed.value))
+		} else if (statusValue & GIT_STATUS_WT_TYPECHANGE.rawValue) == GIT_STATUS_WT_TYPECHANGE.rawValue {
+			status = GitStatus(rawValue: Int(status!.value & GitStatus.workTreeTypeChange.value))
+		}
+
+		return status!
```
mdiep: This should be an `init` on `GitStatus` instead of a method here.

SwiftGit2/Repository.swift
```diff
@@ -537,4 +537,297 @@ final public class Repository {
...
+	}
+
+	private func convertDiffFile(_ file: git_diff_file) -> GitDiffFile {
+		let path = file.path
+		let newFile = GitDiffFile(oid: OID(file.id),
+		                          path: path.map(String.init(cString:))!,
+		                          size: file.size,
+		                          flags: file.flags)
+		return newFile
```
mdiep: This should be an `init` on `GitDiffFile`.

SwiftGit2/Diffs.swift
```diff
@@ -0,0 +1,81 @@
...
+//  Created by Jake Van Alstyne on 8/20/17.
+//  Copyright © 2017 GitHub, Inc. All rights reserved.
+//
+import libgit2
+
+public struct DiffFile {
+	public var oid: OID
+	public var path: String
+	public var size: Int64
+	public var flags: UInt32
```
mdiep: This should be a `DiffFlag`.

SwiftGit2/Diffs.swift
```diff
@@ -0,0 +1,81 @@
...
+	public static let workTreeModified       = Status(rawValue:  1 << 6)
+	public static let workTreeDeleted        = Status(rawValue:  1 << 7)
+	public static let workTreeTypeChange     = Status(rawValue:  1 << 8)
+	public static let workTreeRenamed        = Status(rawValue:  1 << 9)
+	public static let workTreeUnreadable     = Status(rawValue:  1 << 10)
+	public static let ignored                = Status(rawValue:  1 << 11)
+	public static let conflicted             = Status(rawValue:  1 << 12)
+}
+
+public struct DiffFlag: OptionSet {
```
mdiep: I think this should have a plural name. `DiffFlags`

SwiftGit2/Diffs.swift
```diff
@@ -0,0 +1,81 @@
...
+	public var oldFile: DiffFile?
+	public var newFile: DiffFile?
+
+	public init(from diffDelta: git_diff_delta) {
+		self.status = Status(rawValue: diffDelta.status.rawValue)
+		self.flags = DiffFlag(rawValue: diffDelta.flags)
+		self.oldFile = DiffFile(from: diffDelta.old_file)
+		self.newFile = DiffFile(from: diffDelta.new_file)
+	}
+}
```
mdiep: What do you think of adding a `Diff` type that holds the deltas and then nesting these types inside of it?

```swift
public struct Diff {
   public struct File { … }

   public struct Flag: OptionSet { ... }

   public struct Delta {
     public var status: Status
     public var flags: DiffFlag
     public var oldFile: File
     public var newFile: File
   }

   let deltas: [Delta]
}
```

SwiftGit2/Diffs.swift
```diff
@@ -0,0 +1,81 @@
...
+	public static let notBinary  = DiffFlag(rawValue: 1 << 0)
+	public static let validId    = DiffFlag(rawValue: 1 << 1)
+	public static let exists     = DiffFlag(rawValue: 1 << 2)
+}
+
+public struct DiffDelta {
+	public var status: Status?
+	public var flags: DiffFlag?
+	public var oldFile: DiffFile?
+	public var newFile: DiffFile?
```
mdiep: Why are these all optional?

SwiftGit2/Diffs.swift
```diff
@@ -0,0 +1,81 @@
...
+		self.oid = OID(diffFile.id)
+		let path = diffFile.path
+		self.path = path.map(String.init(cString:))!
+		self.size = diffFile.size
+		self.flags = diffFile.flags
+	}
+}
+
+public struct StatusEntry {
+	public var status: Status?
```
mdiep: This probably shouldn't be optional?

SwiftGit2/Repository.swift
```diff
@@ -537,4 +537,142 @@ final public class Repository {
...
+				let parentTreeResult = git_commit_tree(&unsafeParentTree, unwrapParentCommit)
+				guard parentTreeResult == GIT_OK.rawValue, let unwrapParentTree = unsafeParentTree else {
+					return Result.failure(NSError(gitError: parentTreeResult, pointOfFailure: "git_commit_tree"))
+				}
+				git_tree_free(unsafeParentTree)
+
+				var unsafeDiff: OpaquePointer? = nil
+				let diffResult = git_diff_tree_to_tree(&unsafeDiff, self.pointer, unwrapParentTree, unwrapBaseTree, nil)
+				guard diffResult == GIT_OK.rawValue, let unwrapDiffResult = unsafeDiff else {
+					return Result.failure(NSError(gitError: diffResult, pointOfFailure: "git_diff_tree_to_tree"))
```
mdiep: `git_diff_tree_to_tree`, `git_commit_lookup`, and `git_commit_tree` are all used multiple times, but they're quite a bit of code. It seems like we could create helper functions for these? That should really improve readability here.

SwiftGit2/Repository.swift
```diff
@@ -537,4 +537,142 @@ final public class Repository {
...
+
+			if let htoi = s?.pointee.head_to_index {
+				headToIndex = DiffDelta(from: htoi.pointee)
+			}
+
+			if let itow = s?.pointee.index_to_workdir {
+				indexToWorkDir = DiffDelta(from: itow.pointee)
+			}
+
+			let statusEntry = StatusEntry(status: status, headToIndex: headToIndex, indexToWorkDir: indexToWorkDir)
```
mdiep: Can we make the `StatusEntry.init` take the `git_status_entry` and do this work?

mdiep:
Mapping these libgit2 APIs onto Swift is a lot of work! 😅 But you're doing great!

SwiftGit2/Diffs.swift
jakeva: The only caveat with this is that to create an array of a nested struct defined that way, you apparently need to typealias it like so `typealias Delta = Diff.Delta; var returnDict = [Delta]()`, but otherwise i like the organization it provides

SwiftGit2/Diffs.swift
jakeva: you're right, only the files need to be optional as libgit2 will return nil depending on the delta

SwiftGit2Tests/RepositorySpec.swift
```diff
@@ -641,8 +641,63 @@ class RepositorySpec: QuickSpec {
 				expect(commitMessages).to(equal(expectedMessages))
 			}
 		}
+
+		describe("Repository.getRepositoryStatus") {
```
mdiep: This name needs to be updated

SwiftGit2Tests/RepositorySpec.swift
```diff
@@ -641,8 +641,63 @@ class RepositorySpec: QuickSpec {
...
+			}
+
+			it("Should have accurate delta information") {
+				let repo = Fixtures.mantleRepository
+				let branch = repo.localBranch(named: "master").value!
+				expect(repo.checkout(branch, strategy: CheckoutStrategy.None)).to(haveSucceeded())
+
+				let head = repo.HEAD().value!
+				let commit = repo.object(head.oid).value! as! Commit
+				let objects = repo.diff(for: commit)
```
mdiep: This test (and the following ones) are testing `diff`, so they should be in a separate `describe`.

SwiftGit2/Repository.swift
```diff
@@ -537,4 +537,161 @@ final public class Repository {
...
+		/// Get the Base Tree
+		let baseCommit = self.commit(with: commit.oid)
+		guard baseCommit.error == nil else {
+			return Result.failure(baseCommit.error!)
+		}
+
+		let baseTree = self.tree(from: baseCommit.value!)
+		guard baseTree.error == nil else {
+			return Result.failure(baseTree.error!)
+		}
```
mdiep: No that these are nicely abstracted, we can use `flatMap` to propagate errors automatically.

```swift
return self
    .flatMap { baseCommit in
        return self.tree(from: baseCommit)
    }
    .flatMap { baseTree in
        guard !commit.parents.isEmpty else {
            // Initial commit in a repository
            return self.diff(withOldTree: nil, andNewTree: baseTree.value)
        }

        var mergeDiff: Result<OpaquePointer?, NSError> = .success(nil)
        for parent in commit.parents {
            mergeDiff = mergeDiff
                .flatMap { mergeDiff in
                    return self
                        .parentCommit(from: parent)
                        .flatMap { commit in
                            return self.tree(from: commit)
                        }
                        .flatMap { tree in
                            return self.diff(withOldTree: tree, andNewTree: baseTree)
                        }
                        .flatMap { diff in
                            guard let mergeDiff = mergeDiff else { return .success(diff) }
                            let merge = git_diff_merge(mergeDiff, diff)
                            guard mergeResult == gIT_OK.rawValue else {
                                // return return error
                            }
                            return .success(merge)
                        }
                }
        }
    }
    .flatMap { diffResult in
        return self.processDiffDeltas(diffResult)
    }
```

SwiftGit2Tests/RepositorySpec.swift
```diff
@@ -641,8 +641,63 @@ class RepositorySpec: QuickSpec {
...
+
+			it("Should handle initial commit well") {
+				let repo = Fixtures.mantleRepository
+				expect(repo.checkout(OID(string: "047b931bd7f5478340cef5885a6fff713005f4d6")!,
+				                     strategy: CheckoutStrategy.None)).to(haveSucceeded())
+				let head = repo.HEAD().value!
+				let initalCommit = repo.object(head.oid).value! as! Commit
+				let objects = repo.diff(for: initalCommit)
+
+				expect(objects.value?.count).to(equal(2))
```
mdiep: It'd be nice to test the results a bit more fully.

SwiftGit2/Repository.swift
jakeva: interesting. I've seen flatmap before but not like this. I'll put this in

jakeva: Wait, is flatMap supposed to be implemented on Repository? I see it implemented in Result but I'm not sure how to use this sample code.

mdiep: Oops! There was a line missing. I've updated the code sample above.

jakeva:
BTW, I'm noticing Carthage update fails with the Xcode 9 toolchain. I've been looking into why, it appears Nimble doesn't compile and updating the version causes issues with Guanaco… wondering if anyone knows more before I start getting too deep into these other dependencies?

mdiep:
Sorry for the long delay on reviewing. My wife and I adopted a son from China last month. Between the 2 week trip and the adjustments at home, I haven't had as much time for open source recently. But this is on my list and I hope to get to it soon!

jakeva:
No worries! I haven't had much time either, but I think all this needs is more complete test coverage. I haven't looked at it in a while though.

SwiftGit2/Diffs.swift
```diff
@@ -0,0 +1,95 @@
...
+}
+
+public struct Diff {
+	public struct File {
+		public var oid: OID
+		public var path: String
+		public var size: Int64
+		public var flags: Flags
+
+		public init(from diffFile: git_diff_file) {
```
mdiep: This should be an unnamed parameter to match our conventions.

```
public init(_ diffFile: git_diff_file) {
```

SwiftGit2/Diffs.swift
```diff
@@ -0,0 +1,95 @@
...
+		public static let exists     = Flags(rawValue: 1 << 2)
+	}
+
+	public struct Delta {
+		public var status: Status
+		public var flags: Flags
+		public var oldFile: File?
+		public var newFile: File?
+
+		public init(from delta: git_diff_delta) {
```
mdiep: Same here: it should be an unnamed parameter.

SwiftGit2Tests/RepositorySpec.swift
```diff
@@ -641,8 +641,65 @@ class RepositorySpec: QuickSpec {
...
+
+				expect(status.value?.count).to(equal(0))
+
+				let repoWithStatus = Fixtures.sharedInstance.repository(named: "repository-with-status")
+				let branchWithStatus = repoWithStatus.localBranch(named: "master").value!
+				expect(repoWithStatus.checkout(branchWithStatus, strategy: CheckoutStrategy.None)).to(haveSucceeded())
+
+				let statusWithStatus = repoWithStatus.status()
+
+				expect(statusWithStatus.value?.count).to(equal(5))
```
mdiep: This should verify that the actual status are correct—not just verify the total number of statuses. Or there should be separate tests that verify individual statuses.

SwiftGit2Tests/RepositorySpec.swift
```diff
@@ -641,8 +641,65 @@ class RepositorySpec: QuickSpec {
...
+		describe("Repository.status	") {
+			it("Should return accurate status") {
+				// Repo with no status
+				let repo = Fixtures.mantleRepository
+				let branch = repo.localBranch(named: "master").value!
+				expect(repo.checkout(branch, strategy: CheckoutStrategy.None)).to(haveSucceeded())
+
+				let status = repo.status()
+
+				expect(status.value?.count).to(equal(0))
```
mdiep: This `expect` and the one below are really two different tests, so they belong in separate `it`s.

SwiftGit2/Repository.swift
```diff
@@ -537,4 +537,149 @@ final public class Repository {
...
+									guard let mergeDiff = mergeDiff else { return .success(diff) }
+									let mergeResult = git_diff_merge(mergeDiff, diff)
+									guard mergeResult == GIT_OK.rawValue else {
+										return .failure(NSError(gitError: mergeResult, pointOfFailure: "git_diff_merge"))
+									}
+									return .success(mergeDiff)
+							}
+					}
+				}
+				return .success(mergeDiff.value!!)
```
mdiep: I think we can safely get rid of this `!!`.

```swift
			.flatMap { baseTree in
				var mergeDiff: Result<OpaquePointer?, NSError> = .success(nil)
				for parent in commit.parents {
					mergeDiff = mergeDiff
						.flatMap { mergeDiff in
							return self
								.parentCommit(from: parent)
								.flatMap { commit in
									return self.tree(from: commit)
								}
								.flatMap { tree in
									return self.diff(withOldTree: tree, andNewTree: baseTree)
								}
								.flatMap { diff in
									guard let mergeDiff = mergeDiff else { return .success(diff) }
									let mergeResult = git_diff_merge(mergeDiff, diff)
									guard mergeResult == GIT_OK.rawValue else {
										return .failure(NSError(gitError: mergeResult, pointOfFailure: "git_diff_merge"))
									}
									return .success(mergeDiff)
							}
					}
				}
				return mergeDiff.map { ($0, baseTree) }
		        }
			.flatMap { (diffResult, baseTree) in
				if let diffResult = diffResult {
				    return self.processDiffDeltas(diffResult)
				} else {
				    // Initial commit in a repository
				    return self.diff(withOldTree: nil, andNewTree: baseTree)
				}
		}
```

SwiftGit2/Repository.swift
```diff
@@ -537,4 +537,149 @@ final public class Repository {
...
+								}
+								.flatMap { diff in
+									guard let mergeDiff = mergeDiff else { return .success(diff) }
+									let mergeResult = git_diff_merge(mergeDiff, diff)
+									guard mergeResult == GIT_OK.rawValue else {
+										return .failure(NSError(gitError: mergeResult, pointOfFailure: "git_diff_merge"))
+									}
+									return .success(mergeDiff)
+							}
+					}
```
mdiep: We can flatten this a bit by using `Result.fanout`.

```swift
let diff = self
	.parentCommit(from: parent)
	.flatMap { commit in
		return self.tree(from: commit)
	}
	.flatMap { tree in
		return self.diff(withOldTree: tree, andNewTree: baseTree)
	}
return mergeDiff
	.fanout(diff)
	.flatMap { (mergeDiff, diff) in
		guard let mergeDiff = mergeDiff else { return .success(diff) }
		let mergeResult = git_diff_merge(mergeDiff, diff)
		guard mergeResult == GIT_OK.rawValue else {
			return .failure(NSError(gitError: mergeResult, pointOfFailure: "git_diff_merge"))
		}
		return .success(mergeDiff)
	}
```

SwiftGit2/Repository.swift
```diff
@@ -537,4 +537,149 @@ final public class Repository {
...
+		let unsafeParentOid = UnsafeMutablePointer<git_oid>.allocate(capacity: 1)
+		git_oid_fromstr(unsafeParentOid, parent.oid.description)
+		let lookupParentGitResult = git_commit_lookup(&unsafeParentCommit, self.pointer, unsafeParentOid)
+		guard lookupParentGitResult == GIT_OK.rawValue, let unwrapParentCommit = unsafeParentCommit else {
+			return Result.failure(NSError(gitError: lookupParentGitResult, pointOfFailure: "git_commit_lookup"))
+		}
+		git_commit_free(unsafeParentCommit)
+
+		return Result.success(unwrapParentCommit)
+	}
```
mdiep: This whole method should be unneeded since it basically repeats `commit(with: OID)` above.

Instead you should be able to do `commit(with: OID(parent.oid))`.

SwiftGit2/Repository.swift
```diff
@@ -537,4 +537,149 @@ final public class Repository {
...
+		guard lookupBaseGitResult == GIT_OK.rawValue, let unwrapBaseCommit = unsafeBaseCommit else {
+			return Result.failure(NSError(gitError: lookupBaseGitResult, pointOfFailure: "git_commit_lookup"))
+		}
+		git_commit_free(unsafeBaseCommit)
+
+		return Result.success(unwrapBaseCommit)
+	}
+
+	private func diff(withOldTree oldTree: OpaquePointer?,
+	                  andNewTree newTree: OpaquePointer?) -> Result<OpaquePointer, NSError> {
```
mdiep: It doesn't matter that much since this is private, but I think `diff(fromTree:toTree:)` would be more idiomatic labels.

mdiep:
Sorry it took me so long to get back around to this. 😅

SwiftGit2/Repository.swift
jakeva: Removing the !! gives me a strange error I don't understand: `Member 'success' in 'Result<OpaquePointer, NSError>' produces result of type 'Result<T, Error>', but context expects 'Result<OpaquePointer, NSError>'`

SwiftGit2/Repository.swift
jakeva: Good catch :+1:

zdenektopic:
how does it looks with this? I would really use it right now :D

SwiftGit2/Repository.swift
```diff
@@ -537,4 +537,137 @@ final public class Repository {
...
+				for parent in commit.parents {
+					let diff = self
+						.commit(with: parent.oid)
+						.flatMap { commit in
+							return self.tree(from: commit)
+						}
+						.flatMap { tree in
+							return self.diff(fromTree: tree, toTree: baseTree)
+						}
+					return mergeDiff
```
mdiep: This doesn't look correct. If it's a hard `return` here, then it doesn't look like it will work for merge commits (commits with >1 parent).

That means we should add a test for this case.

jakeva:
I'm trying to finish up but I've run into a weird issue. The library builds fine unless I'm building to test, preventing me from running the tests. It fails with 
`
Showing Recent Issues
/Users/jake/Developer/SwiftGit2/Carthage/Checkouts/Nimble/Sources/Nimble/Matchers/AllPass.swift:69:33: Cannot use mutating member on immutable value: 'generator' is a 'let' constant
`

I thought it might have been something wrong with my branch, but I've cloned fresh from `SwiftGit2/SwiftGit2` and still get the problem. Any thoughts? Is this reproducible on your end @mdiep ?

mdiep:
I can reproduce that too. I think the version of Nimble that we're using is incompatible with the latest version of Xcode. That'd make a good separate PR as a precursor to this one. Are you interested in opening a PR to update our dependencies?

jakeva:
I'm looking into it, will let you know what I find.

jakeva:
I've got a question @mdiep that I think might be an easy one for you, but I'm struggling to understand `Result`… 

In the following block what I want to do is check whether mergeDiff is nil. The `guard let` compiles, but does the wrong thing and breaks one of my tests (I believe). Weirdly, doing the straight nil as I have it below check causes the compiler to complain that `Member 'success' in 'Result<OpaquePointer, NSError>' produces result of type 'Result<T, Error>', but context expects 'Result<OpaquePointer, NSError>'`

Anything stand out to you? I would think T would be of type OpaquePointer but Xcode disagrees.

```
return mergeDiff
	.fanout(diff)
	.flatMap { (mergeDiff, diff) in
		if mergeDiff == nil {
			return .success(diff)
		}
//		guard let mergeDiff = mergeDiff else {
//			return .success(diff)
//		}
		let mergeResult = git_diff_merge(mergeDiff, diff)
		guard mergeResult == GIT_OK.rawValue else {
			return .failure(NSError(gitError: mergeResult, pointOfFailure: "git_diff_merge"))
		}
		return .success(mergeDiff)
	}
```

mdiep:
I would stick with the `guard let` that you have.

I think either your test or your expectation is wrong. I think it's the test, because that commit [only changed one file](https://github.com/Mantle/Mantle/commit/d0d9c13da5eb5f9e8cf2a9f1f6ca3bdbe975b57d).

jakeva:
@mdiep It looks like the issue was actually that we were returning mergeDiff too soon, so only one parent was considered. I don't love the way I'm doing it instead (`mergeDiff = mergeDiff.fanout…`) but I can't think of anything else. I've also made the tests more specific, although there is a lot of data not covered such as flags and status values. I started down that road but it started getting a little much.

SwiftGit2/Diffs.swift
```diff
@@ -0,0 +1,95 @@
...
+
+	public init(from statusEntry: git_status_entry) {
+		self.status = Diff.Status(rawValue: statusEntry.status.rawValue)
+
+		if let htoi = statusEntry.head_to_index {
+			self.headToIndex = Diff.Delta(_: htoi.pointee)
+		}
+
+		if let itow = statusEntry.index_to_workdir {
+			self.indexToWorkDir = Diff.Delta(_: itow.pointee)
```
mdiep: You don't need to specify `_: ` here; I didn't even know you could do that. `Diff.Delta(itow.pointee)` is sufficient.

SwiftGit2/Repository.swift
```diff
@@ -558,4 +558,137 @@ final public class Repository {
...
+
+		let count = git_diff_num_deltas(diffResult)
+
+		for i in 0..<count {
+			let delta = git_diff_get_delta(diffResult, i)
+			let gitDiffDelta = Diff.Delta(_: (delta?.pointee)!)
+
+			returnDict.append(gitDiffDelta)
+
+			git_diff_free(OpaquePointer(delta))
```
mdiep: This is incorrect. `git_diff_free` is called on `git_diff`s; `delta` is a `git_diff_delta`.

SwiftGit2/Repository.swift
```diff
@@ -558,4 +558,137 @@ final public class Repository {
...
+		var unsafeBaseCommit: OpaquePointer? = nil
+		let unsafeBaseOid = UnsafeMutablePointer<git_oid>.allocate(capacity: 1)
+		git_oid_fromstr(unsafeBaseOid, oid.description)
+		let lookupBaseGitResult = git_commit_lookup(&unsafeBaseCommit, self.pointer, unsafeBaseOid)
+		guard lookupBaseGitResult == GIT_OK.rawValue, let unwrapBaseCommit = unsafeBaseCommit else {
+			return Result.failure(NSError(gitError: lookupBaseGitResult, pointOfFailure: "git_commit_lookup"))
+		}
+		git_commit_free(unsafeBaseCommit)
+
+		return Result.success(unwrapBaseCommit)
```
mdiep: I don't think it's safe to use `unwrapBaseCommit` here after freeing `unsafeBaseCommit`. I think this should switch to use the same callback style as the `remoteLookup` method.

SwiftGit2/Repository.swift
```diff
@@ -558,4 +558,137 @@ final public class Repository {
...
+
+	private func diff(fromTree oldTree: OpaquePointer?,
+	                  toTree newTree: OpaquePointer?) -> Result<OpaquePointer, NSError> {
+		var unsafeDiff: OpaquePointer? = nil
+		let diffResult = git_diff_tree_to_tree(&unsafeDiff, self.pointer, oldTree, newTree, nil)
+		guard diffResult == GIT_OK.rawValue, let unwrapDiffResult = unsafeDiff else {
+			return Result.failure(NSError(gitError: diffResult, pointOfFailure: "git_diff_tree_to_tree"))
+		}
+
+		return Result.success(unwrapDiffResult)
```
mdiep: Same here about it being unsafe, except this is missing a `git_diff_free`.

SwiftGit2/Repository.swift
```diff
@@ -558,4 +558,137 @@ final public class Repository {
...
+
+	private func tree(from commit: OpaquePointer) -> Result<OpaquePointer, NSError> {
+		var unsafeTree: OpaquePointer? = nil
+		let treeResult = git_commit_tree(&unsafeTree, commit)
+		guard treeResult == GIT_OK.rawValue, let unwrapTree = unsafeTree else {
+			return Result.failure(NSError(gitError: treeResult, pointOfFailure: "git_commit_tree"))
+		}
+		git_tree_free(unsafeTree)
+
+		return Result.success(unwrapTree)
```
mdiep: Same here about it being unsafe.

SwiftGit2/Repository.swift
```diff
@@ -558,4 +558,137 @@ final public class Repository {
...
+
+		let count = git_status_list_entrycount(unwrapStatusResult)
+
+		for i in 0..<count {
+			let s = git_status_byindex(unwrapStatusResult, i)
+			if s?.pointee.status.rawValue == GIT_STATUS_CURRENT.rawValue {
+				continue
+			}
+
+			let statusEntry = StatusEntry(from: (s?.pointee)!)
```
mdiep: Might as well do `s!.pointee` here instead of `(s?.pointee)!`.

SwiftGit2Tests/RepositorySpec.swift
```diff
@@ -645,6 +645,214 @@ class RepositorySpec: QuickSpec {
...
+				let deltas = repo.diff(for: commit).value!
+
+				var newFilePaths: [String] = []
+				for delta in deltas {
+					newFilePaths.append((delta.newFile?.path)!)
+				}
+				var oldFilePaths: [String] = []
+				for delta in deltas {
+					oldFilePaths.append((delta.oldFile?.path)!)
+				}
```
mdiep: These could use `map` instead of `for` loops

```swift
let oldFilePaths = deltas.map { $0.oldFile!.path }
```

mdiep:
I think the tests look good enough now that they include the paths of the files. 👍

SwiftGit2/Diffs.swift
jakeva: nice catch

SwiftGit2/Repository.swift
jakeva: I was working on this when I discovered the already existing `withGitObject`… what do you think about replacing e.g. `.commit(with: parent.oid)` with `.withGitObject(parent.oid, type: GIT_OBJ_COMMIT)` and killing the `commit(with:)` altogether? The tests still pass.

mdiep: Yeah, that sounds good! 👍 You can tell it's been a while since I've really worked with this code. 😅

jakeva: No worries. There's actually a handful of helpers I never noticed before, so I'll convert things to use those.

SwiftGit2/Repository.swift
jakeva: Good catch. According to the docs `The git_diff_delta pointer points to internal data and you do not have to release it when you are done with it.  It will go away when the * git_diff (or any associated git_patch) goes away.`

SwiftGit2/Repository.swift
jakeva: :+1:

SwiftGit2Tests/RepositorySpec.swift
jakeva: nice :+1:

SwiftGit2/Repository.swift
```diff
@@ -561,89 +582,159 @@ final public class Repository {
...
+				return Result<Diff, NSError>.failure(result.error!)
+			}
+			newTree = result.value!
 		}
 
-		let result = Result<OpaquePointer, NSError>.success(tree!)
-		git_tree_free(tree)
-		git_commit_free(commit)
-		return result
+		if oldTree != nil && newTree != nil {
```
jakeva: I'm unsure of this triple branch of logic. I was hoping to find a "swifty" way of identifying which tree was (or whether both trees were) supplied, without needing three `if else` branches but couldn't get it. I'm particularly interested if something pops out, I tried several different things and none were as satisfying to me as just doing the branch.

SwiftGit2/Repository.swift
```diff
@@ -558,4 +579,246 @@ final public class Repository {
...
+				return Result<Diff, NSError>.failure(diff.error!)
+			}
+
+			diffs.append(diff.value!)
+
+			if mergeDiff == nil {
+				mergeDiff = diff.value!
+			} else {
+				let mergeResult = git_diff_merge(mergeDiff, diff.value)
+				guard mergeResult == GIT_OK.rawValue else {
```
mdiep: I think we could just do `git_object_free(diff)` here instead of keeping a list of `diffs`. That would reduce memory consumption and should also make the code a little clearer.

SwiftGit2/Repository.swift
```diff
@@ -558,4 +579,246 @@ final public class Repository {
...
+					return .failure(NSError(gitError: mergeResult, pointOfFailure: "git_diff_merge"))
+				}
+			}
+		}
+
+		for diff in diffs {
+			git_object_free(diff)
+		}
+
+		return .success(Diff(mergeDiff!))
```
mdiep: I think we need to free `mergeDiff`. Putting a `defer { git_object_free(mergeDiff) }` before this line should work.

SwiftGit2/Repository.swift
```diff
@@ -558,4 +579,246 @@ final public class Repository {
...
+		assert(oldCommitOid != nil || newCommitOid != nil, "It is an error to pass nil for both the oldOid and newOid")
+
+		var oldTree: OpaquePointer? = nil
+		if let oid = oldCommitOid {
+			let result = unsafeTreeForCommitId(oid)
+			guard result.error == nil else {
+				return Result.failure(result.error!)
+			}
+
+			oldTree = result.value
```
mdiep: Maybe follow this line with `defer { git_object_free(oldTree) }` (and same with `newTree` below).

The locality of that is nice, but it should also help us free memory if we return an error.

SwiftGit2/Repository.swift
```diff
@@ -558,4 +579,246 @@ final public class Repository {
...
+		}
+
+		for diff in diffs {
+			git_object_free(diff)
+		}
+
+		return .success(Diff(mergeDiff!))
+	}
+
+	/// Caller is responsible to free returned git_diff with git_object_free
```
mdiep: Instead of making the caller free the diff, I think this should take a block that lets you execute code with the diff. That should let us remove the need for the above frees that I comment about.

SwiftGit2/Repository.swift
```diff
@@ -558,4 +579,246 @@ final public class Repository {
...
+		var newTree: Tree? = nil
+		if newCommitOid != nil {
+			let result = self.safeTreeForCommitId(newCommitOid!)
+			guard result.error == nil else {
+				return Result<Diff, NSError>.failure(result.error!)
+			}
+			newTree = result.value!
+		}
+
+		if oldTree != nil && newTree != nil {
```
mdiep: You should be able to replace this with a `switch`.

```swift
switch (oldTree, newTree) {
case let (oldTree?, newTree?): // I'm not sure if you'll be able to reuse the names here
    // has both
case let (tree?, nil), let (nil, tree?):
   // has one
case (nil, nil):
    // has neither
}

File: SwiftGit2/Repository.swift [hunk 45]
```diff
@@ -558,4 +579,246 @@ final public class Repository {
...
+		guard diffResult == GIT_OK.rawValue else {
+			return .failure(NSError(gitError: diffResult,
+															pointOfFailure: "git_diff_tree_to_tree"))
+		}
+
+		return Result<OpaquePointer, NSError>.success(diff!)
+	}
+
+	/// Memory safe
+	private func diff(from oldCommitOid: OID?, to newCommitOid: OID?) -> Result<Diff, NSError> {
```
mdiep: It seems like the safe and un-safe versions of this should share some code.

SwiftGit2/Repository.swift
jakeva: nice catch :+1:

SwiftGit2/Repository.swift
jakeva: :+1:

SwiftGit2/Repository.swift
jakeva: nice!

SwiftGit2/Repository.swift
jakeva: :+1:

SwiftGit2/Repository.swift
jakeva: good call, I might have missed something memory management wise and handling errors was a bit of a trick but I think I have a solution.

SwiftGit2/Repository.swift
jakeva: It looks like `case let (oldTree?, newTree?)` works, the only problem I have is I'm not sure how to use tree here: `case let (tree?, nil), let (nil, tree?):`. I'm not sure how to use this. I need to know which of the two has a value.

mdiep: Ah, it looks like those will need to be separate cases. I didn't see that they were subtly different. Maybe we can reuse some code between them?

SwiftGit2/Repository.swift
```diff
@@ -558,4 +579,234 @@ final public class Repository {
...
+				}
+				return nil
+			}
+
+			if error != nil {
+				return Result<Diff, NSError>.failure(error!)
+			}
+		}
+
+		defer { git_object_free(mergeDiff) }
```
mdiep: This isn't freed in either of the error conditions here.

Let's move this `defer` to be immediately after the `var mergeDiff` declaration.

SwiftGit2/Repository.swift
```diff
@@ -558,4 +579,234 @@ final public class Repository {
...
+
+		var oldTree: OpaquePointer? = nil
+		if let oid = oldCommitOid {
+			let result = unsafeTreeForCommitId(oid)
+			guard result.error == nil else {
+				return transform(Result.failure(result.error!))
+			}
+
+			oldTree = result.value
+			git_object_free(oldTree)
```
mdiep: I think this should be in a `defer` before the `if`?

SwiftGit2/Repository.swift
```diff
@@ -558,4 +579,234 @@ final public class Repository {
...
+
+		var newTree: OpaquePointer? = nil
+		if let oid = newCommitOid {
+			let result = unsafeTreeForCommitId(oid)
+			guard result.error == nil else {
+				return transform(Result.failure(result.error!))
+			}
+
+			newTree = result.value
+			git_object_free(newTree)
```
mdiep: I think this should be in a `defer` before the `if`?

SwiftGit2/Repository.swift
```diff
@@ -225,6 +225,27 @@ final public class Repository {
...
 
+	private func withGitObjects<T>(_ oids: [OID], type: git_otype, transform: ([OpaquePointer]) -> Result<T, NSError>) -> Result<T, NSError> {
+		var pointers = [OpaquePointer]()
+		for oid in oids {
+			var pointer: OpaquePointer? = nil
+			var oid = oid.oid
+			let result = git_object_lookup(&pointer, self.pointer, &oid, type)
+
+			guard result == GIT_OK.rawValue else {
+				return Result.failure(NSError(gitError: result, pointOfFailure: "git_object_lookup"))
```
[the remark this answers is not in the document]
jakeva: I think the point of collecting them was to allow the caller to perform custom behavior on their values e.g. on line 683. The collection is then freed in 243-245. Or did you mean something else?

SwiftGit2/Repository.swift
jakeva: good catch :+1:

SwiftGit2/Repository.swift
jakeva: It's gotchas like these that help me learn. I'm a rails developer for my day job. Thanks!

SwiftGit2/Repository.swift
jakeva: 👍

jakeva:
I'd be happy to see about adding to the documentation, where can I find that?

SwiftGit2/Repository.swift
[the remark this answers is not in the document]
mdiep: I mean that in the case where we return an error here, they need to be freed.

mdiep:
> I'd be happy to see about adding to the documentation, where can I find that?

Any documentation should be added inline with 3 `/`s.

```swift
/// Information about this struct
struct Struct {
    /// Information about this method
    ///
    /// - parameters:
    ///    - foo: Information about foo
    ///  - returns: Information about the return value
    func doSomething(foo: String) -> Int {
    }
}
```

SwiftGit2/Repository.swift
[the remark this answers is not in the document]
jakeva: Got it. How about another `defer`?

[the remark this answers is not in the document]
mdiep: Oh yeah! For some reason I was thinking that wouldn't work but it totally does. 🤦‍♂️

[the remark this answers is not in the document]
jakeva: 👍

SwiftGit2/Repository.swift
```diff
@@ -558,4 +582,234 @@ final public class Repository {
...
+		let pointer = UnsafeMutablePointer<git_status_options>.allocate(capacity: 1)
+		let optionsResult = git_status_init_options(pointer, UInt32(GIT_STATUS_OPTIONS_VERSION))
+		guard optionsResult == GIT_OK.rawValue else {
+			return .failure(NSError(gitError: optionsResult, pointOfFailure: "git_status_init_options"))
+		}
+		var options = pointer.move()
+		pointer.deallocate(capacity: 1)
+
+		var unsafeStatus: OpaquePointer? = nil
+		let statusResult = git_status_list_new(&unsafeStatus, self.pointer, &options)
```
mdiep: One last memory issue: This needs a `git_status_list_free` in a `defer`.

mdiep:
After this last fix we can merge! 🙈

SwiftGit2/Repository.swift
jakeva: 👍

jakeva:
Sounds good! Thanks for finding all those memory leaks! I should have been more diligent about considering that.

mdiep:
No worries! I think I missed them several times too. 🙈 

Thanks for all your work on this! Sorry to put you through so many rounds of review.

jakeva:
Not at all! glad we got it right, and you've helped me learn a lot. 👍  💯

mdiep:
:tada:

PR #104, opened 2017-10-24
Switch to just one target (no more -iOS -OSX)

AndrewSB:
I loosely followed http://promisekit.org/news/2016/08/Multiplatform-Single-Scheme-Xcode-Projects/, I built the new target on iOS & watchOS

This makes using swift-package-manager much easier


mdiep:
How does this relate to SwiftPM?

I'm generally against universal frameworks, but I'm willing to reconsider if there's a compelling reason.

AndrewSB:
I should have added more context, my apologies. This relates to SwiftPM in that it's easier to maintain one target instead of 4 in `Package.swift`. 

Would you care to share why you're against universal frameworks? If this change will negatively effect the framework we should abandon it. It's only value add is that it can simplify maintenance.

For even more context, you can look at our conversation about implementing the same thing for Moya at https://github.com/Moya/Moya/issues/818 and https://github.com/Moya/Moya/pull/967. 967 has review notes and potential downsides that may be relevant in our case

mdiep:
We should be able to have 4 schemes here and 1 target in Package.swift. The two are unrelated. (See https://github.com/ReactiveCocoa/ReactiveSwift/ for example.)



> Would you care to share why you're against universal frameworks? If this change will negatively effect the framework we should abandon it. It's only value add is that it can simplify maintenance.

I haven't found that universal frameworks simplify maintenance. Having everything separate makes it easier to understand what exactly is going on. But it can be hard to tell what build settings apply to which platforms or tell if there's any interference.

If the tooling (i.e. Xcode) were designed to use universal frameworks, it would be a different matter.

AndrewSB:
Interesting, I'll take a look.

I see, so it's like I'm working against the tool. I'll try to incorporate
that into my mindset. Great feedback, thanks Matt 🙃

I'll go ahead and close and change #105 so it doesn't depend on the change

PR #105, opened 2017-10-24
Swift package manager support!

AndrewSB:
Depends on #104


mdiep:
Thanks for the PRs! I'll get to this and your other PR as soon as I can, but I have a backlog of PRs to review. 😅

SwiftGit2/SwiftGit2.swift
```diff
@@ -0,0 +1,14 @@
...
+//  SwiftGit2
+//
+//  Created by Andrew Breckenridge on 10/23/17.
+//  MIT.
+//
+
+import libgit2
+
+// swiftlint:disable:next identifier_name
+static func SwiftGit2Init(_: Void) {
```
mdiep: This isn't equivalent to what was in the Obj-C version. The `__attribute__((constructor))` causes the function to run when the library is loaded. We'd need some sort of equivalent mechanism here. (It could probably be replaced with a well-placed `dispatch_once`.)

Package.swift
```diff
@@ -0,0 +1,26 @@
...
+    dependencies: [
+	  .package(url: "https://github.com/antitypical/Result.git", "3.0.0" ..< "4.0.0"),
+    ],
+    targets: [
+        // Targets are the basic building blocks of a package. A target can define a module or a test suite.
+        // Targets can depend on other targets in this package, and on products in packages which this package depends on.
+        .target(
+            name: "SwiftGit2",
+            dependencies: ["Result"],
+	    path: "SwiftGit2"),
```
mdiep: Formatting seems off here.

mdiep:
Thanks for the PR! It'd be great to support SwiftPM. ✨ 

Could you also add SwiftPM to CI when addressing the feedback above?

AndrewSB:
Not an issue! I'm probably going to take at least a couple weeks to get back around to this. If any other contributor would like to help, I'd take a PR on my fork to finish up spm support.

Also, in my opinion it makes more sense to add CI in a separate PR, but if you're sure you want it this way I'll oblige

mdiep:
> Also, in my opinion it makes more sense to add CI in a separate PR, but if you're sure you want it this way I'll oblige

I like adding it here so that we can have a good test that it works.

SwiftGit2/SwiftGit2.swift
AndrewSB: Great catch. In swift this should just be possible with a `static let`, I remember reading that it accomplishes the same result as the `dispatch_once` dance in Obj-C. Can we write a test for this?

mdiep: Looking at the documentation for `git_libgit2_init`, we might be able to just call this in `init` and `dealloc` of `Repository`.

mdiep: I'm not sure how to write a test for it. CI wasn't able to build with your changes. Do the tests successfully run without calling this?

AndrewSB: I wasn't sure either, let's circle back once CI for SwiftPM is up

AndrewSB:
Anyone else who sees this is welcome to contribute as well! 

## TODO before merge

- [ ] remove dependency on #104 
- [ ] Add CI support
- [ ] remove `static func SwiftGit2Init(_: Void)` and call from `init` and `dealloc` of `Repository`

Issue #103, opened 2017-08-28
Switch to using a Command type


mdiep:
Instead of putting everything as a method on `Repository`, it might be beneficial to reify the methods into a `Command` type.

```swift
struct Command<Value, Error> {
    func flatMap<U>(transform: (Value) -> Command<U, Error>) -> Command<U, Error>
    func map<U>(transform: (Value) -> U) -> Command<U, Error>
}

class Repository {
    func execute<V, E>(_ command: Command<V, E>) -> Result<V, E>
}
```

One benefit of this would be colocating methods with the types that they operate on. e.g, instead of `Repository.allRemotes() -> Result<[Remote], NSError>`, we could have `Remote.all() -> Command<[Command], NSError>`. That could make code more readable.

It also opens the door to composition. Just as you can write your own `git` command that combines built-in commands, you could write your own `Command` that combines multiple `Command`s.

Issue #74, opened 2016-12-22
Custom Error type


mattrubin:
https://github.com/SwiftGit2/SwiftGit2/pull/72#discussion_r93550202

DivineDominion:
As a first step, what do you think about replacing `NSError` with `Result.AnyError`? Then you can at least retrieve the original `error` without any loss of information and not that much work involved. Later one can replace that with more expressive custom error types.

DivineDominion:
Nevermind, that doesn't help at all because the underlying error codes are just numbers 🙃

PR #111, opened 2017-12-15
Remote fetching

chriseidhof:
I wanted to take a stab at fetching a remote. A bunch of things that you might not like (open for suggestions):

- I changed `Remote` to be class (it needs to manage the pointer)
- The `name` / `URL` properties don't copy anymore, but are backed by the libgit2 calls
- I use `deinit` for memory management
- I duplicate the pointer instead of deleting a free call somewhere else
- ~~This is based off of the fix-build branch~~

### DIFF
--- a/SwiftGit2/Repository.swift
+++ b/SwiftGit2/Repository.swift
@@ -345,22 +345,43 @@ final public class Repository {
 		return Result.success(remotes.map { $0.value! })
 	}
 
+	private func remoteLookup<A>(named name: String, _ callback: (Result<OpaquePointer, NSError>) -> A) -> A {
+		var pointer: OpaquePointer? = nil
+		let result = git_remote_lookup(&pointer, self.pointer, name)
+
+		guard result == GIT_OK.rawValue else {
+			return callback(.failure(NSError(gitError: result, pointOfFailure: "git_remote_lookup")))
+		}
+
+		defer { git_remote_free(pointer) }
+		return callback(.success(pointer!))
+	}
+
 	/// Load a remote from the repository.
 	///
 	/// name - The name of the remote.
 	///
 	/// Returns the remote if it exists, or an error.
 	public func remote(named name: String) -> Result<Remote, NSError> {
-		var pointer: OpaquePointer? = nil
-		let result = git_remote_lookup(&pointer, self.pointer, name)
+		return remoteLookup(named: name) { $0.map(Remote.init) }
+	}
 
-		guard result == GIT_OK.rawValue else {
-			return Result.failure(NSError(gitError: result, pointOfFailure: "git_remote_lookup"))
+	/// Download new data and update tips
+	public func fetch(_ remote: Remote) -> Result<(), NSError> {
+		return remoteLookup(named: remote.name) { remote in
+			remote.flatMap { pointer in
+				var opts = git_fetch_options()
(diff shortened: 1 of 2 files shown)


mdiep:
Thanks for opening a PR! 😄 

You're right that I don't like the things you listed. 🙈

My [design](https://github.com/SwiftGit2/SwiftGit2#design) for SwiftGit2 is based off my experience with https://github.com/libgit2/objective-git in GitHub for Mac. The Objective-Git is similar to what you have here.

That was the source of many headaches because:

1. You need to think more actively about concurrency. Libgit2 access must be serial within an open repository. This introduces a lot of cognitive overhead to application development.

1. What looks like a simple property access may be quite slow. It's hard to anticipate performance implications when looking at the code or when running under ideal circumstances. (I have a similar dislike for faulting ORMs like Core Data or ActiveRecord.)

1. You introduce co-effects all over the place instead of containing them, which makes code hard to test.

So I do think it'd be better to add a `Repository.fetch` method that looks up the remote by name and then fetches. (Although I do hope to move to some sort of command API #103)

chriseidhof:
Understood, that makes sense. I see how that would work, will change my PR.

chriseidhof:
I've changed the code to your suggestions, what do you think?

SwiftGit2/Repository.swift
```diff
@@ -362,6 +362,27 @@ final public class Repository {
 		git_remote_free(pointer)
 		return Result.success(value)
 	}
+	
+	public func fetch(_ remote: Remote) -> Result<(), NSError> {
+		var pointer: OpaquePointer? = nil
+		let result0 = git_remote_lookup(&pointer, self.pointer, remote.name)
+		guard result0 == GIT_OK.rawValue else {
```
chriseidhof: I don't really like `result0` and `result`, but I don't know if we should assume the lookup succeeds (maybe the remote has been deleted in the meantime?)

mdiep: I think you're right to not assume that the lookup succeeds. I think I've elsewhere named the variable something like `lookupResult`. It would probably be ideal to return a more specific error (c.f. #74) and use an enum case for when the remote no longer exists, but I don't think that's required for now.

We could probably add a helper that's used here and in `remote(named:)` above that actually performed the lookup and gives you access to the pointer for the remote inside a block. If it also freed the remote, that would probably also be helpful—it looks like you're missing a `git_remote_free` here.

Or add a type that wraps `OpaquePointer` with a freeing function to call in `deinit`. Then the helper could return a `Result<SmartPointer, NSError>` and this method could use `flatMap`. 🤷‍♂️

chriseidhof: Yes. So the think I like a lot about a class-based approach with deinit is that you don't need to think so much about memory management: you always pass in a pointer and free it in deinit. I think having some kind of system for this would be nice.

chriseidhof: (also: good catch). I just pushed some new changes and a helper.

mdiep:
Looks good. 👍 

There's just a conflict that needs to be resolved because I removed trailing whitespace in #112.

chriseidhof:
I rebased off of master and force-pushed.

mdiep:
🎉

chriseidhof:
Yay!

PR #120, opened 2018-03-26
Fix mismatched status types

utahwithak:
Fixes this issue:
https://github.com/SwiftGit2/SwiftGit2/issues/119

Uses the libgit2 values to ensure no future problems arise.


mdiep:
This is great! Thanks for opening a PR. This approach is way better.

PR #122, opened 2018-03-27
Swift Package Manager

stonehouse:
Forked this PR #105 and continued working on it. I've got it so that ```swift build``` actually works, though it's using the system libgit2. You could hypothetically link to a static binary but I haven't gotten that working. I still need to get tests working (via swift test) but I need to get some of the test dependencies using Swift PM as well. Just thought I'd create the PR now and get your feedback. 

I initially just had the init call in ```Repository.init``` but it turned out to be necessary in some of the class funcs.

I've also renamed libgit2 to 'Clibgit' which is the recommended style by Swift PM. If you guys have an issue with that I can change it back, or potentially make that consistent throughout the project.


mdiep:
Thanks for working on this!

> I've also renamed libgit2 to 'Clibgit' which is the recommended style by Swift PM.

Can you point me to that recommendation? I wasn't aware of that.

I'm definitely not against doing that. I'd prefer to follow any recommendations. But:

1. It should probably be `Clibgit2`
2. I'd definitely want to make it consistent throughout the project

Package.resolved
```diff
@@ -0,0 +1,25 @@
+{
+  "object": {
+    "pins": [
+      {
+        "package": "Clibgit",
+        "repositoryURL": "git@github.com:stonehouse/Clibgit.git",
```
mdiep: Let's include this in the SwiftGit2 repository instead of using an external dependency.

SwiftGit2/SwiftGit2.swift
```diff
@@ -0,0 +1,18 @@
...
+#if SWIFT_PACKAGE
+	import Clibgit
+#else
+	import libgit2
+#endif
+
+// swiftlint:disable:next identifier_name
+let SwiftGit2Init: () = {
+    git_libgit2_init()
+}()
```
mdiep: Nice!

stonehouse:
Here's the documentation referring to C library naming conventions.
https://github.com/apple/swift-package-manager/blob/master/Documentation/Usage.md#require-system-libraries

Here's the excerpt
```
The convention we hope the community will adopt is to prefix such modules with C and to camelcase the modules as per Swift module name conventions. Then the community is free to name another module simply libgit which contains more “Swifty” function wrappers around the raw C interface.
```

I'm not so sure about naming repos like this simply "libgit2" but I do like the idea of prepending "C" to make it clear that this is literally just a C API.

Package.resolved
stonehouse: So, with Swift Package Manager it needs a separate repo to define the dependency. It's kind of weird now, rather than system modules being just another product, it is at the root of the package. I'm happy for you to fork my repo and then the dependency is owned by the same account. But as far as I can tell, you need a separate git repo to wrap the C binary. It's actually discussed in the section of the docs that I linked below.

mdiep:
Ha! That's funny that they use libgit2 as an example. I think `Clibgit2` is a better idea than `Clibgit` since the `2` is part of the name. But 👍 to the `C` prefix.

stonehouse:
Everything now uses "Clibgit2" as the module name. There seems to be a crash in the iOS tests finding the repo "repository-with-status". Is that a regression or was that already in master?

So, something I'd still like to do is get the tests running in Swift PM as well, but I think I need to get some of the dependencies to use Swift PM first. Apart from that everything should work, there's just the question of the external dependency. Maybe there's a way to include it in the repo, but it seems like SwiftPM is very much based on having separate repos for dependencies.

mdiep:
> So, something I'd still like to do is get the tests running in Swift PM as well, but I think I need to get some of the dependencies to use Swift PM first. Apart from that everything should work, there's just the question of the external dependency. Maybe there's a way to include it in the repo, but it seems like SwiftPM is very much based on having separate repos for dependencies.

I think Carthage's [`Package.swift`](https://github.com/Carthage/Carthage/blob/master/Package.swift) might shed some light on this. You don't define it as a dependency but as a target within the package AIUI.

stonehouse:
All the targets in Carthage are proper targets that have source files to be build. If you look at all those target folders they all have swift source files. Swift PM does support C/C++/Obj-C code as well, but again it needs to build all those. I haven't looked how complicated the libgit2 build is, but I imagine it's not trivial to get it building with Swift PM. What I've done is just make a system module package, which just links against a libgit2.dylib found in the system.

_Things I've tried_
1. Include a pkgConfig property in SwiftGit2, telling it to include libgit2.
```error: configuration of package 'SwiftGit2' is invalid; the 'pkgConfig' property can only be used with a System Module Package```
2. Create a 'Clibgit2' target in the SwiftGit2 Package.
```warning: target 'Clibgit2' in package 'SwiftGit2' contains no valid source files```

It seems like it's a limitation and or design choice currently that you need a separate repo. We could just have two separate "packages", but that's not currently supported either.

**Comment by stonehouse** [comment]
On a related note, if I try and link against the static libgit2 binaries in the External folder then I get the following error. Any idea what that is? With the current build you have to have the libgit2 dylib on your system.

```
Undefined symbols for architecture x86_64:
  "_iconv", referenced from:
      _git_path_iconv in libgit2.a(path.c.o)
     (maybe you meant: _git_path_iconv_init_precompose, _git_path_iconv , _git_path_iconv_clear )
  "_iconv_close", referenced from:
      _git_path_iconv_clear in libgit2.a(path.c.o)
  "_iconv_open", referenced from:
      _git_path_iconv_init_precompose in libgit2.a(path.c.o)
ld: symbol(s) not found for architecture x86_64
<unknown>:0: error: link command failed with exit code 1 (use -v to see invocation)
```

**Comment by mdiep** [comment]
Sorry for the delay on this. 😞 

The one-Package-per-repository limitation makes sense. We can keep it in a separate repo. When we're ready to merge this, I can create a `Carthage/Clibgit2` repository to host it.

> With the current build you have to have the libgit2 dylib on your system.

I can't tell this is in response to trying to move `Clibgit2` into this repo or if it's a problem that would prevent us from merging this. 😕

File: Package.resolved [hunk 3]
```diff
@@ -0,0 +1,34 @@
...
+        "state": {
+          "branch": null,
+          "revision": "5cc96d3222624a6b1e23dbaabaadbac284766643",
+          "version": "1.0.0"
+        }
```
mdiep: This `Clibgit` version needs to be removed.

mdiep: Meant to request changes, not approve. 🙈

mdiep:
> I can create a SwiftGit2/Clibgit2 repository to host it.

I've gone ahead [and created this](https://github.com/SwiftGit2/Clibgit2). You can open a PR there with the package contents.

twobitunicorn:
What is the scoop with this pull request?  I am working on Swift Package support now and it would be dreamy if this got in.

stonehouse:
Hi @twobitunicorn sorry I never got back to this. My PR probably needs a rebase, but it does work with SwiftPM. What doesn't work currently are the tests (via SwiftPM) because some of the test dependencies don't support SwiftPM. Apart from that, the Clibgit2 C-Module is hosted on my Github profile, I think it needs to be moved to /SwiftGit2. I used this with SwiftPM for a while myself, but I saw some random crashes performing some actions. The dilemma is that SwiftPM only really has support for including system libraries, and doesn't support referencing a random framework, so the Clibgit2 repo just points to the location of the system framework. So, effectively you're using a different libgit2 binary than you would via CocoaPods.

sja26:
@stonehouse I've fixed the SPM Tests issue and created a [pull request]( https://github.com/stonehouse/SwiftGit2/pull/2).

BilalReffas:
Feel free to reopen if after you working on this again.

PR #123, opened 2018-04-14
[WIP] Swift 4.1

DivineDominion:
I have trouble running the macOS tests, so this is not mergeable. But it's a start.

Things to discuss that I'd then change:

- [ ] Changing `deallocate(capacity: 1)` to `deallocate()` is a backwards-compatibility breaking change; should I wrap that in `#if swift` version checks instead?
- [ ] build an run the OSX tests (doesn't work for me because `@rpath/libswiftAppKit.dylib` cannot be loaded)
- [ ] fix SwiftLint version, e.g. by checking it out via CocoaPods and invoking the local version [like ReSwift does](https://github.com/ReSwift/ReSwift)

### DIFF
--- a/SwiftGit2/OID.swift
+++ b/SwiftGit2/OID.swift
@@ -27,12 +27,12 @@ public struct OID {
 		let result = git_oid_fromstr(pointer, string)
 
 		if result < GIT_OK.rawValue {
-			pointer.deallocate(capacity: 1)
+			pointer.deallocate()
 			return nil
 		}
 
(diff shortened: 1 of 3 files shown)


DivineDominion:
@mdiep can you run the macOS tests without problems? 🤔

mdiep:
I am able to run the macOS tests without issues.

DivineDominion:
Oh since a Swift 4.1 update was pushed in the past, we can forget about this one :)

PR #126, opened 2018-04-30
update to Swift 4

tengyifei:
@mdiep This is the first of a bunch of PRs!

========================================================

Travis does not recognize 9.3?

fix swift 4.1 deprecations

fix more swift 4.1 deprecation

use Xcode 9.3 in CI

PR #127, opened 2018-05-01
Git Add

tengyifei:
@mdiep This is the second PR after Swift 4.1 implementing `git add` by path. Once #126 is merged I can rebase this on top of master.


SwiftGit2/Repository.swift
```diff
@@ -583,6 +583,33 @@ final public class Repository {
 		return iterator
 	}
 
+	/// Get the index for the repo. The caller is responsible for freeing the index.
+	public func unsafeIndex() -> Result<OpaquePointer, NSError> {
```
mdiep: Let's not make this `public`. Unsafe operation should be kept internal.

mdiep:
This looks good apart from the one note above!

tengyifei:
@mdiep made that change!

PR #136, opened 2018-09-26
add Repository.isValid

DivineDominion:
Adds a boolean check that some libgit2 bindings have, too, to avoid the overhead of dealing with the real "open" return values.


mdiep:
Looks great! Thanks!

PR #137, opened 2018-09-26
Update Project to Swift 4.2

DivineDominion:
Not much to do here

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: objective-c
 
 os: osx
-osx_image: xcode9.3
+osx_image: xcode10.1
 
 sudo: false # Enable container-based builds
 
--- a/Cartfile
+++ b/Cartfile
@@ -1,1 +1,1 @@
-github "antitypical/Result" ~> 3.0
+github "antitypical/Result"
--- a/Cartfile.resolved
+++ b/Cartfile.resolved
@@ -1,5 +1,5 @@
-github "Quick/Nimble" "v7.0.3"
-github "Quick/Quick" "v1.2.0"
+github "Quick/Nimble" "v7.3.1"
+github "Quick/Quick" "v1.3.2"
 github "ZipArchive/ZipArchive" "v1.8.1"
 github "antitypical/Result" "3.2.4"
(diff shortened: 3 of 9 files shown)


mdiep:
Unfortunately the macOS build failed. 😞 

We might need to move to a newer Travis image.

arguiot:
@DivineDominion I've sent a PR to your forked repo that updates Result, so it's compatible with Swift 4.2

DivineDominion:
@mdiep I updated the PR accordingly.

mdiep:
```
⚠️  /Users/travis/build/SwiftGit2/SwiftGit2/External/libgit2/src/unix/posix.h:86:9: 'futimens' is only available on macOS 10.13 or newer [-Wunguarded-availability-new]
        return futimens(f, s);
```

It looks like we may need to bump up the deployment target?

DivineDominion:
I think I give up on fixing Travis build in my life at this point :) Always a puzzle. You're right, the latest job still fails, even though Xcode 10.1 is used and macOS 10.13 as well (as [per the xcode-setting](https://docs.travis-ci.com/user/reference/osx/#macos-version), not the os-setting in the YML config file, of course ... ... ... (insert disappointed stare here) ...):
https://travis-ci.org/SwiftGit2/SwiftGit2/jobs/515265488

The log is too long to be displayed, by the way; someone with more Travis experience than I should do something about that :)

Not sure what this is about and why my local, freshly checked-out build works. macOS 10.13 is used by Travis, so ¯\\_(ツ)_/¯

The only thing I know that should affect the build is to use `osx_image: xcode10.2` which will make Travis compile on 10.14

mdiep:
I have no problem jumping straight to Swift 5

DivineDominion:
Ok. Building ZipArchive causes trouble in the time being unless this is merged: https://github.com/ZipArchive/ZipArchive/pull/500

mdiep:
Looks like that was merged and release. 😃

DivineDominion:
This is like detective work, only I am both detective and the suspect leaving trails :) Will be working on a Swift 5 branch now :)

PR #149, opened 2019-04-12
Simplify Equatable and Hashable implementations

mattrubin:
- For types where the manual `==` and/or `hash(into:)` implementations were equivalent to what the compiler would synthesize, the explicit implementations have been deleted.
- For protocols with common implementations for all types conforming to the protocol, the implementations have been unified in a protocol extension.
- Several `==`implementations have been converted from free functions to static methods.


mdiep:
Awesome! It's a relief to get all these cleaned up. Brings back memories of how things used to be. 😄

PR #150, opened 2019-04-13
Default Sequence methods on CommitIterator

mattrubin:
The standard library offers default implementations of many Sequence-related methods via protocol extensions. This PR switches `CommitIterator` to use those implementations instead of its own. This is particularly useful for the seven methods that `CommitIterator` exposed publicly but did not properly implement.


mdiep:
👏🏻

PR #158, opened 2019-07-03
Swift Package Manager support without changing dependency.

muizidn:
The urge to support Swift Package Manager comes from the project which added SwiftGit2 as the dependency through SwiftPM.

But there are some problems since current SPM doesn't support managing build configurations in a file. It provides passing flags during invocation which is very cumbersome.

Using Makefile like approach doesn't scale well since the project that uses SwiftGit2 as a dependency will need to implement every single configuration in their project.

Current approach, we can use simple script which reads configuration files and computes it to be passed in SwiftPM invocation. This is utility offered by [SPMCLI](https://github.com/pondok-programmer/spmcli). It helps define configuration in a file and resolves all config in the project so that the maintainer and users are free from rewriting configurations defined in dependecy. SPMCLI also store resolved config so it can be examined later. Note, this is temporary solution until SPM provides nicer way.

Is there any other solution?
Yes, there are some solution. You can use Makefile, Rakefile, Fb buck, Google bazel. They are more powerful but also have disadvantages.

Disadvantages using above approach is we need Python installation and SPMCLI in our $PATH.

### Test
For now the test configuration has been defined in SPMCLI althoug `spmcli test` invocation throws error durint the test. That's a problem to fix, but the xcodeproj test is the remain success.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+.vscode/
+.build/
+
 # Xcode
 #
 build/
--- a/SwiftGit2/CheckoutStrategy.swift
+++ b/SwiftGit2/CheckoutStrategy.swift
@@ -6,7 +6,7 @@
 //  Copyright (c) 2015 GitHub, Inc. All rights reserved.
 //
 
-import libgit2
+import Clibgit2
 
 /// The flags defining how a checkout should be performed.
 /// More detail is available in the libgit2 documentation for `git_checkout_strategy_t`.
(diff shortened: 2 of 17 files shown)


mdiep:
Thanks for the PR! I'd love to get this merged, but the number of changes and files make it hard to review this. Could you first open a PR that renames `libgit2` to `Clibgit2`? That would make reviewing a lot more straightforward.

muizidn:
Ahh. Ok.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

muizidn:
@mdiep I'm ready to go.

mdiep:
This looks good. Thanks!

PR #159, opened 2019-07-24
Support for Swift Package Manager

muizidn:
This PR is continuation of #158

Changes in this PR including
1. Update libgit2 directory to Clibgit2 to match SwiftPM rule and update Xcodeproj setting to point new updated path.
2. Add Package.swift file. To automate task, we need to incorporate SPMCLI files.

SwiftPM doesn't support mixed source package. Hence, we rename SwiftPackage2/*.swift files to SwiftPackage2/Swift/*.swift.

PR #170, opened 2020-04-28
SwiftPM Support

BilalReffas:
This will add following PR from @muizidn [Support for Swift Package Manager #159](https://github.com/SwiftGit2/SwiftGit2/pull/159)

It has been already reviewed by multiple people.


alexanderwe:
Would really like to see this land. This would simplify the integration into command line tools made with SPM

kareman:
I don't think SPM is currently capable of compiling libgit2, so I'm working on a solution at https://github.com/kareman/SwiftGit2/tree/spm-binary-target that uses a precompiled framework, but it will only work in Swift 5.3 and later. Which is currently in beta.

I have gotten it to compile and run, but the new support for resources seems incomplete. I'm working on a workaround right now.

alexanderwe:
@kareman It sounds nice to have libgit2 compiled with SPM. Looking forward to it for Swift 5.3.

But having the opportunity to specify SwiftGit2 in an Package.swift, would already be a first step.

BilalReffas:
Hi Folks,

I didn't forget this. Will work on this soon again, just to much things at work currently. Anyway my first attempt failed as it was more a hack.

@kareman this sounds as a reliable solution. I would prefer definitely this.

We can vendor libgit2 and build it as part of your package in our repository in order to make this work, though.

But again was @kareman is doing sounds better IMHO.

kareman:
The PR for the precompiled approach I mentioned is #172

BilalReffas:
Will take a look into this for reviewing. Thank you so much for submitting this.

PR #193, opened 2021-06-02
Fix Mac ARM framework creation

DivineDominion:
This fixes compilation for Mac targets using Carthage. I would rather we use https://github.com/light-tech/LibGit2-On-iOS/issues/1 and XCFrameworks eventually, but this works in my projects so far.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ DerivedData
 *.idea*
 
 External/libgit2*.a
+External/libgit2-mac
 External/ios-openssl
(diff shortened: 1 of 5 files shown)


DivineDominion:
@mdiep Just upgraded the script to handle Xcode 13+ correctly, so that ARM target are build with Xcode 12.2 and newer.

DivineDominion:
@SwiftGit2/swiftgit2 (hope that ping works :)) What do y'all think about merging this to master?

Issue #197, opened 2022-02-10
Build SwiftGit2 on Apple Silicon


hannesa2:
How to make it work on a x86 and an arm64 mac ?
With an Homebrew arm64 the file is located at `/opt/homebrew/lib/libssh2.a`

![image](https://user-images.githubusercontent.com/3314607/153345133-738e4f1f-2370-4178-95f1-74fd5ffc6d06.png)

hannesa2:
Because you have no arm64 runner, I used my custom arm64 runner https://github.com/hannesa2/SwiftGit2/pull/2 to have a smoke run.

But I run into 
<img width="1086" alt="image" src="https://github.com/SwiftGit2/SwiftGit2/assets/3314607/a2acd580-af49-4431-828b-9dcb8c158a99">

DivineDominion:
@hannesa2 Could you check out `xcode-15` locally and maybe tweak your runners? I'm too inexperienced and slow with that, so I don't feel confident in getting that right. But it builds for my apps locally when I use that branch, and the binary also runs on all platform's I've tested.

hannesa2:
Official Github runner only have Xcode 14 , please see https://github.com/hannesa2/SwiftGit2/actions/runs/6688925628/job/18171685874

Run ls -la /Applications/Xcode*
lrwxr-xr-x  1 runner  admin  28 Oct 17 23:06 /Applications/Xcode.app -> /Applications/Xcode_14.2.app
lrwxr-xr-x  1 runner  admin  28 Oct 17 23:00 /Applications/Xcode_13.1.0.app -> /Applications/Xcode_13.1.app
lrwxr-xr-x  1 runner  admin  30 Oct 17 23:00 /Applications/Xcode_13.2.app -> /Applications/Xcode_13.2.1.app
lrwxr-xr-x  1 runner  admin  30 Oct 17 23:00 /Applications/Xcode_13.3.app -> /Applications/Xcode_13.3.1.app
lrwxr-xr-x  1 runner  admin  30 Oct 17 23:00 /Applications/Xcode_13.4.app -> /Applications/Xcode_13.4.1.app
lrwxr-xr-x  1 runner  admin  30 Oct 17 23:00 /Applications/Xcode_14.0.app -> /Applications/Xcode_14.0.1.app
lrwxr-xr-x  1 runner  admin  28 Oct 17 23:00 /Applications/Xcode_14.1.0.app -> /Applications/Xcode_14.1.app
[...]

PR #201, opened 2023-10-09
Github action on pull request

hannesa2:
Travis is no more used since a lot of time here.
And I want to help (inspired by https://github.com/SwiftGit2/SwiftGit2/issues/197#issuecomment-1753254757) to have at least a CI build similar as I did here https://github.com/gitx/gitx

Any improvement is welcome and currently `script/cibuild` doesn't work properly.
That why I started with `SwiftGit2-OSX` build. -step by step-

If all is working I can share my arm64 runner as well.
Please keep in mind, I'm no Apple habitat developer.

### DIFF
--- a/.github/workflows/BuildPR.yml
+++ b/.github/workflows/BuildPR.yml
@@ -1,0 +1,41 @@
+name: pull request
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        xcode: [ Xcode_14.1.0, Xcode_13.3 ]
+        os: [ macos-12 ]
+        abi: [ x86_64 ]
+#        include:
+#          - xcode: Xcode
+#            os: ARM64
+#            abi: arm64
+    steps:
+      - name: ls Xcode
+        run: ls -la /Applications/Xcode*
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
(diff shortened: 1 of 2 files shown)


hannesa2:
i see, a build from a fork doesn't build as long as it's not on upstream.
But here you can see how does it looks like https://github.com/hannesa2/SwiftGit2/pull/1

.github/workflows/BuildPR.yml
```diff
@@ -0,0 +1,41 @@
...
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        xcode: [ Xcode_14.1.0, Xcode_13.3 ]
+        os: [ macos-12 ]
+        abi: [ x86_64 ]
```
fwcd, line 17: Nit: This isn't the ABI, this is the architecture (the ABI would e.g. be `gnu` in `x86_64-unknown-linux-gnu`, on macOS there's only one standard ABI though, so it's usually omitted). See [the Clang docs](https://clang.llvm.org/docs/CrossCompilation.html#target-triple) for details.

hannesa2: Ahh, this is copy&paste leftover from other repo. 
Here I remove it https://github.com/SwiftGit2/SwiftGit2/pull/206

hannesa2: ok, the name is misleading

hannesa2: https://github.com/SwiftGit2/SwiftGit2/pull/207

PR #204, opened 2023-10-17
WIP: Build iOS

hannesa2:


hannesa2:
> @hannesa2 Is this superseded by #205?

I would say: The first job which build properly wins

mbernson:
Hi,

I'm working on a git GUI client for Mac as a hobby project, using libgit2 and a fork of SwiftGit2. My aim is to leverage the Swift Package Manager as much as possible. Because despite what the name implies, SPM also support packages that consist of C, C++ or Objective-C code.

I've written a Package.swift file for libgit2 that makes it compile cleanly on Apple platforms. Perhaps it is of use to you:
https://github.com/libgit2/libgit2/compare/main...mbernson:libgit2:spm

As you probably know, libgit2 needs transport 'backends' for HTTPS and SSH, so that it can communicate with a remote git server.

### HTTPS

At first, I compiled OpenSSL from source as an xcframework, and added it as a dependency to the aforementioned Package.swift file. ([Gist on how to do that](https://gist.github.com/mbernson/8e09fa53de0fa1c942b0a26f6f0e8c7c)). That works. But using OpenSSL means that you need to ship your a bundle of trusted certificates with your app yourself.

Instead, it's also possible to have libgit2 use Apple's networking framework 'SecureTransport', just by defining `GIT_SECURE_TRANSPORT=1`. That's what I'm using now.

### SSH

I've managed to compile libssh2 for Mac ARM64 as an xcframework, but haven't tried it yet. For now, I've defined `GIT_SSH_EXEC=1` in my package, so that libgit2 shells out to OpenSSH. I've yet to decide if I'll keep that.

---

That's all I've got for now - if people want it and my time permits it, I can make a PR to contribute some of these changes from my fork to the upstream SwiftGit2!

DivineDominion:
@mbernson That sounds really good. A PR would be great to get rid of all the weirdness of the old build system, eventually. If you eventually find the time to look into this, it'd be much appreciated!

DivineDominion:
@hannesa2 Turns out neither won :)

PR #205, opened 2023-10-18
Build all with script

hannesa2:
It builds all like the travis one did

PR #203, opened 2023-10-17
WIP: Xcode 15 build

DivineDominion:

PR #187, opened 2021-01-28
Add SPM, Linux support, GitHub Actions and perform some cleanup

fwcd:
### Fixes #167 and #95

This PR is based on #172 and replaces the Carthage/Xcode project with a minimal, cross-platform layout that follows the SwiftPM conventions and integrates `libgit2` through an (external) `.systemLibrary` target. The advantage of this approach is that it is now easy to build and test the package on platforms like Linux with minimal setup:

```
apt install libgit2-dev
swift test
```

To verify this, it also adds a GitHub Actions workflow that performs a full build/test run on Ubuntu 20.04.

PR #208, opened 2024-03-02
Migrate SwiftGit2 to a Swift Package

mbernson:
This PR migrates SwiftGit2 to a Swift Package. This means we can get rid of the Xcode project file, build scripts and Carthage configuration. 

I’ve updated the GitHub Actions workflow. As you can see in [the GitHub Actions run on my fork](https://github.com/mbernson/SwiftGit2/actions/runs/11403645255), the library compiles without warnings and the test suite passes on a matrix of: macOS 14, macOS 15 and iOS 17.2.

Related to pull requests #187, #203, #204, #205.

# Breaking changes

The following changes are breaking for users of SwiftGit2:

- The library must now be explicitly initialized by the application using the library using `SwiftGit2.initialize()`, and may be shut down using `SwfitGit2.shutdown()`.
  - Previously, SwiftGit2 used `__attribute__((constructor))` to automatically call this when the library is loaded ([code](https://github.com/SwiftGit2/SwiftGit2/blob/5eb09593a0cbeb65aaf3cca1591901d4b639b1ef/SwiftGit2/SwiftGit2.m#L13-L15)). But I'm not aware of any way to do this in pure Swift.
- Removed Xcode projects, Carthage files, and build scripts. If anyone is using SwiftGit2 via Carthage, this will break their workflow.

# Other changes

These changes have no effect for users of SwiftGit2:

- The project now uses the standard Swift package structure, so the source files have moved around a bit.
- Update tests to use the latest versions of Nimble and Quick.

# Missing features

**SSH and HTTPS**: I haven’t added LibSSH2 and OpenSSL back yet. They can be compiled as XCFrameworks, and added as binary dependencies to the SwiftGit2 package.

I've figured out how to compile OpenSSL in this way ([code](https://gist.github.com/mbernson/8e09fa53de0fa1c942b0a26f6f0e8c7c)) and how to compile LibSSH2 ([code](https://gist.github.com/mbernson/e830f34d1bcd88b8e7a97d21ca9dce24)). These could be built using GitHub Actions. But I don't know where we would host the XCFramework files for distribution.


mbernson:
Tagging @DivineDominion and @hannesa2 since you were interested in this!

kaunteya:
Hi. Looking forward to getting this branch merged

stevengharris:
This is a great change to SwiftGit2, thanks. FWIW, I added `.macCatalyst(.v15)` to platforms and it works beautifully also.

mbernson:
Thank you! I've noticed some other folks who have starred/forked my repo in order to use SwifGit2 via SPM, or who have reached out to me via e-mail. So there is some demand for this!

I'm still willing to help and ship it. The plan of action would be:
1. Create a fork of libgit2 under the SwiftGit2 GitHub organisation, which adds the [`Package.swift` file](https://github.com/mbernson/libgit2/blob/spm/Package.swift).
2. Update the Package.swift file in this PR to point to it.
3. Merge this PR.
4. Release a new version of SwiftGit2, bumping the major version number (since there are backwards-incompatible changes).

Follow up steps would be:
* Create repos under the SwiftGit2 organisation that automatically build OpenSSL and libssh2 as xcframeworks, which are served on GitHub Packages. That way, SwiftGit2 can connect to remotes :)

.github/workflows/BuildPR.yml
```diff
@@ -8,34 +8,24 @@ on:
...
-      fail-fast: false
-      matrix:
-        xcode: [ Xcode_14.1.0, Xcode_13.3 ]
-        os: [ macos-12 ]
-        architecture: [ x86_64 ]
-#        include:
-#          - xcode: Xcode
-#            os: ARM64
-#            architecture: arm64
+    runs-on: macos-14
```
hannesa2, line 11: Why do you removed the matrix ?

Btw, now you run an arm64 runner 👍, but I guess it makes sense to test on x86 as well.

hannesa2 (commented):
About the swift I can't contribute, I only have some pipeline knowledge, and here I would suggest to test arm64 and x86

.github/workflows/BuildPR.yml
mbernson: I've added back the CI matrix configuration. Note that on Github's macOS 15 image only Xcode 16 is available, so I've only specified that Xcode version right now.

mbernson:
It's been a while since I had time to work on this. Today I made the following changes:
1. Added libgit2 as a git submodule to this repo. This is easier to manage than the libgit2 fork that I had.
2. Added Mac Catalyst platform to Package.swift manifest.
3. Added back the CI matrix configuration. Tests are passing [over on my fork](https://github.com/mbernson/SwiftGit2/pull/1).

DivineDominion (approved):
Thank you! 🙏  This is very cool.

I finally made some time to check out your branch and run some tests. It appears that everything works nicely -- and in my app it's even *faster* to build than the Carthage-based approach of yesteryear.

Could we lower to macOS 10.13 (would be somewhat selfish reasons, I admit; but I'm fine with a local fork)?

DivineDominion:
Summoning SwiftGit2 owners @BilalReffas @mdiep -- I can't open a new repo to upload the framework binaries to. That idea sounds sensible if we can pull it off!

@mbernson Maye we can merge not into `master` but a parallel branch until this is fully solved to not break feature compatibility fully?

feifanzhou:
@mbernson thank you for doing this! Looking forward to this PR getting merged

mbernson:
Thanks for your reply @BilalReffas. It's been a while again since I worked on this PR, since it is very much a side project for me. 😬 I'd be happy to help out as a co-maintainer. I do have enough time to review other people's PR's.

I would love to merge this, and then we should also increment the major version number to indicate a backwards-incompatible change. I've seen the recent PR activity on this project. That means that some people still depend on the Carthage-based workflow.

Sources/SwiftGit2/SwiftGit2.swift
```diff
@@ -0,0 +1,53 @@
...
+//  SwiftGit2.swift
+//
+//
+//  Created by Mathijs Bernson on 01/03/2024.
+//
+
+import Foundation
+import Clibgit2
+
+public enum SwiftGit2 {
```
mbernson, line 11: I've discovered a problem with this approach to namespacing. It makes it no longer possible to reference a type in SwiftGit2 using its fully qualified name (such as `SwiftGit2.Tag`).

I will change these into freestanding functions to avoid that problem.

stevengharris:
FWIW, I was using @mbernson's fork for a while and moved over to use this PR directly a couple of weeks ago. It's been working without issues for me, and I'm using it in an app that walks across what I think would be a pretty good cross section of the API.

DivineDominion:
Phew, ok so 3/5 CI job passed -- sorry for the noise. I don't understand why those that fail do fail, yet, but we should be able to fix that and then merge with a new major version bump.

mbernson:
The `build-ios` job is running these commands:
```
DEVICE=`xcrun xctrace list devices 2>&1 | grep -oE 'iPhone.*?[^\(]+' | head -1 | awk '{$1=$1;print}' | sed -e "s/ Simulator$//"`
xcodebuild -scheme SwiftGit2 -destination "platform=$PLATFORM,name=$DEVICE" build test
```
Which becomes:
```
/Applications/Xcode_16.3.app/Contents/Developer/usr/bin/xcodebuild -scheme SwiftGit2 -destination "platform=iOS Simulator,name=iPhone 15" build test
```
And fails with the message "The requested device could not be found because no available devices matched the request."

I've actually run into this very same problem on the iOS projects that I'm developing for my day job. The strangest thing is that the exact same CI workflow passes on our self-hosted GitHub Actions runners, but fails with the error message above on the GitHub provided runners.

Some options to fix it would be:
1. Create the Simulator in the workflow using `xcrun simctl create`
2. Use a resilient custom action like https://github.com/mxcl/xcodebuild

mbernson:
The error in the `build-macos` job appears to be a C compiler error; super strange... I can reproduce it locally on my M4 Mac running macOS 15 and Xcode 16. I will try to look into it some more.

mbernson:
Ok, I've figured out the compilation issue on macOS 15 with Xcode 16.

It was caused by the zlib dependency that is bundled with libgit2. This particular version of zlib fails to compile on Clang 17 with the error we've just seen. This problem was fixed in libgit2 v1.9.0 in which they updated zlib.

However, libgit2 v1.9.0 introduced a new compilation issue: an unterminated multiline comment, which is fixed in this commit which is not part of any libgit2 release yet:
https://github.com/libgit2/libgit2/commit/2fa13adf09c8dd1f334f57023390043ea3f71cb2

When I check out the libgit2 submodule at that commit (and remove a reference to a deprecated constant from SwiftGit2) it is solved.

I'm surprised that both of these syntax errors were ever shipped in libgit2 and not caught in any testing. 😅 Anyway, I'll push a commit that fixes it.

mbernson:
@DivineDominion I've also addressed the other build failure by using the [xcodebuild action](https://github.com/mxcl/xcodebuild) (by Max Howell, the Homebrew creator) instead of kicking off `xcodebuild` by hand. This has an added benefit that it's now much easier to specify a test matrix of iOS versions to test on!

All of the jobs are now passing. ✅ (I know this because the same workflow runs on my fork of SwiftGit2: https://github.com/mbernson/SwiftGit2/actions/runs/14750305831)

mbernson:
@DivineDominion I've fixed the build, could you have another look at this?

DivineDominion:
@mbernson I am making time and am looking into the changes this week. I verified that a local copy with Xcode 26 also compiles fine. 🚀 

Because there's a multi-step release process of breaking changes, I believe we're fine merging to `master` as long as we're not creating a release just yet. I don't see any other fixes for anything floating around that this would get in the way of.

I don't have SwiftGit2 org access, though. @BilalReffas would need to prepare the changes to host the .xcframeworks 

Because of that, I'd rather not merge to master, but a feature branch instead, so that we have wiggle room to work in parallel. We could even create pre-release tags from that for an upcoming major version, skipping 1.0 :)

I created `release-2.0` so we all have a reference branch to work on and would like to ask you to target that instead.

Could you mention breaking changes and the current limitations in the README in a section at the very start since this is WIP? Then we don't need to reference this PR after the merge.

DivineDominion:
About hosting the frameworks -- if all else fails, there's apparently potential to stuff it into the git repo: https://github.com/joehinkle11/SwiftGit3/blob/8382c67472bfb159c980a0a65bb41968dd612b8f/Package.swift

mbernson:
@DivineDominion That sounds like a plan! I've re-pointed my PR to the `release-2.0` branch and have added a notice at the top of the README as you've suggested.

I'm confident that we'll find a good solution for the XCFrameworks!

I also have another branch ready that converts all the unit tests to the new Swift Testing framework, which I intend to submit as a PR after this one. :)

mbernson:
I think we're in the clear to merge this to the `release-2.0` branch now?

I already have another PR lined up that converts all the tests to the Swift Testing framework. With all tests passing on my fork of SwiftGit2. 😄 https://github.com/mbernson/SwiftGit2/pull/3

mbernson:
Hi all, 
I'm currently on my Christmas break from work and am spending some time on my side-project again which uses SwiftGit2.  I wanted to share some progress on integrating the dependencies OpenSSL and libssh2.

In short: I've successfully integrated OpenSSL and libssh2 with SwiftGit2 via SPM and can now do git fetches! This unlocks all of SwiftGit2's capabilities that it had before moving to SPM.

### How it works

I have two GitHub Actions workflows up and running:
* [Build OpenSSL XCFramework](https://github.com/mbernson/OpenSSL-Apple/blob/8f2b7a1c43bb1bacc211d8ad43bdf3c2a407a496/.github/workflows/build_openssl.yml)
* [Build libssh2 XCFramework](https://github.com/mbernson/libssh2-Apple/blob/ca5759ffbe54a39c2a33ed2f88cced0e4b128c55/.github/workflows/build_libssh2.yml)

I've downloaded the compiled frameworks and added them locally for testing. In `Package.swift`:
```swift
// ...
targets: [
    .target(
        name: "Clibgit2",
        dependencies: ["OpenSSL", "libssh2"],
        // ...
    ),
    .binaryTarget(name: "OpenSSL", path: "path/to/OpenSSL.xcframework"),
    .binaryTarget(name: "libssh2", path: "path/to/libssh2.xcframework"),
]
```

Then, also in `Package.swift` we tell libgit2 to use these libraries as the SSH and HTTPS backends:
```diff
// Git SSH transport configuration
.define("GIT_SSH", to: "1"),
-.define("GIT_SSH_EXEC", to: "1"),
+.define("GIT_SSH_LIBSSH2", to: "1"), // Use libssh2

// Git HTTPS transport configuration
.define("GIT_HTTPS", to: "1"),
.define("GIT_HTTPPARSER_BUILTIN", to: "1"),
-.define("GIT_SECURE_TRANSPORT", to: "1"),
+.define("GIT_OPENSSL", to: "1"), // Use OpenSSL
```

I had to define a helper function so I can tell libgit2 where to find the macOS trusted TLS certificates. Unfortunately, this needs to be done in C, because the `git_libgit2_opts` API uses variadic arguments which cannot be bridged to Swift. 
```c
#include <git2.h>

int git_opt_set_ssl_cert_locations(const char *file, const char *path) {
    return git_libgit2_opts(GIT_OPT_SET_SSL_CERT_LOCATIONS, file, path);
}
```
And calling it:
```swift
git_opt_set_ssl_cert_locations("/etc/ssl/cert.pem", nil)
```

With these changes, clone and fetch operations over HTTPS now use OpenSSL instead of Apple's [SecureTransport](https://developer.apple.com/documentation/security/secure-transport) (which I now notice is deprecated).

Fetching over SSH also works. I've tested this using the "in memory" credentials provider, but the other methods should also work after dealing with the macOS sandboxing restrictions.

### Next steps

Checking compiled frameworks in to the SwiftGit2 repo would be a bad practice, so we should host the frameworks on a public HTTPS url. I think the best option for this is to use GitHub Releases. 

The GitHub Actions workflows I've shared above live in their own repositories. These could automatically:
1. Build the xcframeworks for macOS, iOS and any other supported Apple platforms.
2. Create a tagged GitHub Release and attach the binaries. 
We can then reference these in SwiftGit2.

hannesa2:
What prevents a merge here ?

PR #210, opened 2024-08-12
MacOS 13

hannesa2:
It's also somehow a preparation for an arm64 runner

### DIFF
--- a/script/update_libgit2
+++ b/script/update_libgit2
@@ -5,18 +5,19 @@ set -e
 # augment path to help it find cmake installed in /usr/local/bin,
 # e.g. via brew. Xcode's Run Script phase doesn't seem to honor
 # ~/.MacOSX/environment.plist
-PATH="/usr/local/bin:$PATH"
+if [[ `uname -m` == 'arm64' ]]; then
+  echo "** Running on a Apple Silicon M1"
+  PATH="/usr/homebrew/bin:$PATH"
+else
+  echo "** Running on a Apple x86"
+  PATH="/usr/local/bin:$PATH"
+fi
 
 SCRIPT_DIR=$(dirname "$0")
 source "${SCRIPT_DIR}/xcode_functions.sh"
 
 function setup_build_environment ()
 {
-    # augment path to help it find cmake installed in /usr/local/bin,
-    # e.g. via brew. Xcode's Run Script phase doesn't seem to honor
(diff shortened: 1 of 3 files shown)


script/update_libgit2
```diff
@@ -5,18 +5,19 @@ set -e
 # augment path to help it find cmake installed in /usr/local/bin,
 # e.g. via brew. Xcode's Run Script phase doesn't seem to honor
 # ~/.MacOSX/environment.plist
-PATH="/usr/local/bin:$PATH"
+if [[ `uname -m` == 'arm64' ]]; then
+  echo "** Running on a Apple Silicon M1"
+  PATH="/usr/homebrew/bin:$PATH"
```
DivineDominion, line 10: IIRC it's `/opt/homebrew/bin` now by default, per https://brew.sh/ samples and:

```
❯ which cmake
/opt/homebrew/bin/cmake
```

DivineDominion (changes requested):
Thanks for the changes! I believe the Homebrew path needs to be adjusted; Could you add an ARM architecture to the test matrix and check? Otherwise this looks good 👍

script/update_libgit2
hannesa2: @DivineDominion  done

DivineDominion:
@hannesa2 Thank you for the quick reaction!  You mentioned in the SPM issue that you're experienced with pipelines -- can we get GitHub CI to run on both Intel and arm64 to verify the PATH setting without too much trouble? To me it looks like we can't do that within this test matrix because we need different images,  i.e. `macos-13-xlarge`:
https://github.com/actions/runner-images?tab=readme-ov-file#available-images

But I'm not sure myself tbh

hannesa2:
Yes I could do, but step by step.
I want to avoid to add more and more features to one pull request, which is never merged. (and focus an interest is lost)
So please merge _first_ this one, _than_ I can jump inst arm64 issue with an new pull request ....

DivineDominion:
@hannesa2 Fair! While this PATH change may break, the old PATH wouldn't have worked, either, so let's do it this way.

DivineDominion:
FWIW my Xcode 15 fixes from this PR still worked locally: https://github.com/SwiftGit2/SwiftGit2/pull/203/files#diff-745c5da16ef62f9f0fe570838efbdeebd58735407d9fef130b38749d5ee3d2f8

PR #212, opened 2024-11-18
Use an arm64 runner for CI

hannesa2:
Let's see where the issues are 
image

### DIFF
--- a/.github/workflows/BuildPR.yml
+++ b/.github/workflows/BuildPR.yml
@@ -14,14 +14,11 @@ jobs:
       matrix:
         xcode: [ Xcode_14.3, Xcode_15.2.0 ]
         os: [ macos-13 ]
-        architecture: [ x86_64 ]
         include:
           - xcode: Xcode_15.4
             os: macos-14
-            architecture: arm64
           - xcode: Xcode_16
             os: macos-14
-            architecture: arm64
     steps:
       - name: ls Xcode
         run: ls -la /Applications/Xcode*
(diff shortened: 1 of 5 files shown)


hannesa2:
@DivineDominion 
Now it builds on arm64 as well.

Next upcoming thing could be a release pipeline, with an autogenerated changelog

.github/workflows/BuildPR.yml
```diff
@@ -15,10 +15,13 @@ jobs:
...
-#          - xcode: Xcode
-#            os: ARM64
-#            architecture: arm64
+        include:
+          - xcode: Xcode_15.4
+            os: macos-14
+            architecture: arm64
+          - xcode: Xcode_16
+            os: macos-14
+            architecture: arm64
```
DivineDominion, line 24: Sorry for asking: do these `include` declarations supersede the 3 lines above that have an array of xcode, os, and architecture? If so, can these 3 lines above be removed?

DivineDominion:
Thank you for powering through this! Just have one question real quick to maybe remove redundant instructions. But we're good to merge either way.

DivineDominion (approved):

.github/workflows/BuildPR.yml
hannesa2: No it adds to x86 matrix

hannesa2:
I solved the conflict

.github/workflows/BuildPR.yml
hannesa2: A normal matrix is a multiplication, so you can't simple add arm64. This must be added to matrix separated

DivineDominion:
Sorry for the delay! With our toddler at home, I just didn't notice that so much time has passed. Thank you for your work on this!

